PR #1, opened 2016-05-12
fix(actions): Wasn't actually emitting the correct actions to the actions Subject

jayphelps:
Was instead accidentally emitting only async actions, which is useless.

```diff
if (typeof action === 'function') {
    let obs = from(action(actions, store));
    let sub = obs.subscribe(next);
-   actions.next(action);
    return sub;
  } else {
+   actions.next(action);
    return next(action);
  }
```


blesh:
Merged with a1cf32e

PR #6, opened 2016-05-13
chore(publish): don't publish all the things, fixes #5

### LINKED ISSUE #5: Shouldn't publish the resources directory to npm
Right now we're publishing the images and everything in the resources directory to npm. That's unnecessary

jayphelps:


README.md
```diff
@@ -1,4 +1,5 @@
-![LOGO](resources/logo-medium.png)
+![LOGO](https://cloud.githubusercontent.com/assets/762949/15234285/e84cc032-1864-11e6-9fb5-ced867249343.png)
```
jayphelps: This way, we don't need to relatively reference the logo if someone opens it locally without `resources` directory.

package.json
```diff
@@ -10,6 +10,11 @@
     "test": "npm run build && npm run build_tests && mocha temp",
     "prepublish": "npm test"
   },
+  "files": [
```
jayphelps: whitelist instead of blacklist is easier to maintain IMO.

PR #18, opened 2016-05-14
Adds TypeScript typings, fixes #12

### LINKED ISSUE #12: Add TypeScript type definitions
Do you plan to add typings for this library?

bennett000:
Here's an attempt at providing typings for (#12).  There are a number of different ways of doing typings, I took the approach that seemed easiest.  I would be happy to modify the typings into alternate styles, or configurations as I made a lot of assumptions. 

- adds a new `index.d.ts` file which has the typings
- adds a `typings` property in `package.json` which points to `index.d.ts`
- adds a `test/typings.ts` file which attempts to use the typings
- adds an `npm run clean` task to remove `lib` and `temp`
- adds an `npm run test_typings` task to encapsulate the typescript test command
- adds `rimraf` and `typescript` as `devDependencies`
- updates `.gitignore` to exclude intelliJ files
- modifies `npm run test` so that it executes the TypeScript test

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,25 @@
+import { Middleware } from 'redux';
+import { Observable } from 'rxjs/Observable';
+import { Operator } from 'rxjs/Operator';
+
+export declare function reduxObservable(): Middleware;
+
+// ./node_modules/rxjs/Observable.d.ts
+export declare class ActionsObservable<T> extends Observable<T> {
+
+  // ./node_modules/rxjs/Observable.d.ts specifies an `Observable`'s `source`
+  // property like so: `protected source: Observable<any>;`
+  // since `actionsSubject` is being assigned to an `Observable`'s `source`
+  // property it is being given the type: `Observable<any>`
+  constructor(actionsSubject: Observable<any>);
+
+  // ./node_modules/rxjs/Observable.d.ts specifies operators as generics like so
+  // `protected operator: Operator<any, T>;`
+  lift(operator: Operator<any, T>);
+
+  // ./node_modules/redux/index.d.ts specifies `Action` as having a `type`
+  // property with a type signature of `any`
+  // since `key` is being compared with an `Action`'s `type`, `key` has a type
+  // signature of `any`
+  ofType(key: any);
+}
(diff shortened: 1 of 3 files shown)


bennett000:
Looking at the diff I think my editor expanded part of `package.json` for no reason.  I can fix that if you like.

admmasters:
The only thing I would add - is potentially there is some noise in the commit that may not want to be part of this PR - some of the formatting of the contributors in the package.json and the gitignore file etc. Ultimately, the typings look good to me though.

bennett000:
@admmasters I fixed the contributors noise and reverted the `.gitignore`.  I can take out the clean command too if you want better atomicity.

index.d.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  // ./node_modules/rxjs/Observable.d.ts specifies operators as generics like so
+  // `protected operator: Operator<any, T>;`
+  lift(operator: Operator<any, T>);
+
+  // ./node_modules/redux/index.d.ts specifies `Action` as having a `type`
+  // property with a type signature of `any`
+  // since `key` is being compared with an `Action`'s `type`, `key` has a type
+  // signature of `any`
+  ofType(key: any);
```
jayphelps: Shouldn't this be a `string`? Anything else they would pass would be a bug.

bennett000: @jayphelps that would have been my instinct as well.  When the official typings were added to Redux there was discussion around whether `type` should be a string, or some sort of union: https://github.com/reactjs/redux/pull/1413#issuecomment-185560527 The conversation eventually ends with the decision to use `any` for `type`: https://github.com/reactjs/redux/pull/1413#issuecomment-186652132

All that being said I do not have a strong opinion but it "feels right" to follow the main library's convention.

jayphelps: ohhh I hadn't seen that discussion. `any` it is. While I think redux should have taken a stronger stance, we should follow their lead. Thank you for be so thorough!

bennett000: I was just lucky to have been following the Redux typings scene for the last few months.

jayphelps:
thank you

index.d.ts
admmasters: Thanks for this!

PR #19, opened 2016-05-14
feat(reduxObservable): allow async streams to emit other async actions

### LINKED ISSUE #8: Add support for Observable emitting additional "thunkservables"?
Okay... so I just coined a term "thunkservable"... which is a function that returns an observable.

Anyhow, should we support something like this?

```js
let secondAction = () => Observable.of({ type: 'ACTIONS_DONE' });
let firstAction = () => Observable.interval(1000).mapTo(secondAction);

dispatch(firstAction);
```

jayphelps:
Fixes #8 

Basically, instead of the middleware just passing emitted actions to _the next middleware_ we instead dispatch it to the store, which allows async action streams to emit other async actions (aka thunkservables), as well as allows recursive behavior via the `ActionsObservable`.

### DIFF
--- a/src/reduxObservable.js
+++ b/src/reduxObservable.js
@@ -10,7 +10,7 @@ export function reduxObservable() {
     return (action) => {
       if (typeof action === 'function') {
         let obs = from(action(actionsObs, store));
-        let sub = obs.subscribe(next);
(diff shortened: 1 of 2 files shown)


blesh:
:shipit: LGTM

Issue #11, opened 2016-05-13
Best practice question with a flatmap-debounced-search example 


winkerVSbecks:
Thank you for this library – it's awesome!

I was trying to test out a more complex example. Something that replicates a search input which makes a debounced api call onChange, etc.

This is what I came up with, it's a complete example so you can clone and run it:
https://github.com/winkerVSbecks/redux-obsv-search-example/blob/master/src/actions/index.js#L10

Is this the right way to go about it? My primary concern is whether you need that init search action or can this somehow be achieved using only one onChange action.

jayphelps:
We have been thinking about this too. #9 had an idea that didn't pan out.

We've been playing with a couple approaches, one similar to yours, and another dispatching an action but including a stream of search terms.

http://jsbin.com/ziwuya/edit?js,output

```js
const fetchSearchResults = (searchTerms) => (
  (actions) => (
    searchTerms
      .switchMap(searchTerm =>
        ajax('/api/search', searchTerm)
          .map(
            payload => ({ type: 'RESULTS_FULLFILLED', payload })
        )
      )
  )
);
```
```js
const SearchResults = ({ searchResults, fetchSearchResults }) => (
  <div>
    <input
      placeholder="Enter search term..."
      ref={input =>
        fetchSearchResults(
          Rx.Observable.fromEvent(input, 'input')
            .map(event => event.target.value)
            .debounceTime(200)
        ) 
      }
    />
    <ul>
      {searchResults.map(result =>
        <li key={result}>{result}</li>
      )}
    </ul>
  </div>
);

export default connect(
  ({ searchResults }) => ({ searchResults }),
  dispatch => ({
    fetchSearchResults: q => dispatch(fetchSearchResults(q))
  })
)(SearchResults);
```

We're not sure yet which of these two (or another) is the best approach.

jayphelps:
I've updated my above example and the jsbin to handle unsubscribe teardown (and renamed `fetchSearchResults` to `streamSearchResults` to more accurately describe what it is doing)

winkerVSbecks:
That's interesting. With this approach it almost feels like there is a parallel architecture to redux – not sure how I feel about that. It would be nice to come up with an approach to limit any side-effects/creating of observables to the action creators.

jayphelps:
@winkerVSbecks See #38

PR #26, opened 2016-05-17
feat(dispatchOnMount): a decorator to manage subscriptions over the c…

blesh:
…omponent lifecycle

Adds a higher-order-component decorator that will dispatch multiple actions and thunkservables on componentDidMount, then manage any subscriptions that are returned by thunkservables and ensure they are disposed of when componentWillUnmount is called.

related #11

### DIFF
--- a/src/dispatchOnMount.js
+++ b/src/dispatchOnMount.js
@@ -1,0 +1,23 @@
+import { Component } from 'react';
+import { Subscription } from 'rxjs/Subscription';
+
+const $$reduxObservableSubscription = '@@reduxObservableSubscription';
+
+export function dispatchOnMount(...toDispatch) {
+  return (ComposedComponent) =>
+    class DispatchOnMountComponent extends Component {
+      componentDidMount() {
+        this[$$reduxObservableSubscription] = new Subscription();
+        toDispatch.map(a => this.context.store.dispatch(a))
+          .forEach(sub => sub && this[$$reduxObservableSubscription].add(sub));
+      }
+
+      componentWillUnmount() {
+        this[$$reduxObservableSubscription].unsubscribe();
+      }
+
+      render() {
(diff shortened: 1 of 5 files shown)


blesh:
This does add a peer dependency on react, unfortunately. We might want to consider moving this to a separate package so support things like Angular 2.

cc/ @jayphelps

blesh:
Usage is as we discussed, @jayphelps. We'll need to document it heavily.

jayphelps:
Shoot I forgot about the peer dependency thing. How bout we create `react-redux-observable` that we can stuff these sorts of helpers in?

src/dispatchOnMount.js
```diff
@@ -0,0 +1,23 @@
...
+import { Subscription } from 'rxjs/Subscription';
+
+const $$reduxObservableSubscription = '@@reduxObservableSubscription';
+
+export function dispatchOnMount(...toDispatch) {
+  return (ComposedComponent) =>
+    class DispatchOnMountComponent extends Component {
+      componentDidMount() {
+        this[$$reduxObservableSubscription] = new Subscription();
+        toDispatch.map(a => this.context.store.dispatch(a))
```
jayphelps: How does this work without
```js
static contextTypes = {
  store: PropTypes.shape({
    dispatch: PropTypes.func.isRequired,
  })
};
```
I thought React required you to provide `contextTypes` to opt-in to receiving a context, otherwise it's not provided?? Am I drunk?

The docs even say so https://facebook.github.io/react/docs/context.html#passing-info-automatically-through-a-tree
> If contextTypes is not defined, then this.context will be an empty object.

blesh: Yep... bad test. Need to fix.

blesh:
Closed in favor of [react-redux-observable](/redux-observable/react-redux-observable)

PR #28, opened 2016-05-19
docs(examples): Add clickable `localhost` link

sotojuan:
I'm lazy and I want to be able to click a link in my terminal (iTerm 2) when I run an example. This makes it so I can do that.


jayphelps:
Thanks! I am lazy also.

Issue #25, opened 2016-05-17
Add optional default scheduler argument to middleware factory


blesh:
The idea is provide the option to schedule observation with a scheduler passed to the middleware factory. This might prove useful for integration testing and possibly other uses.

PR #30, opened 2016-05-19
Add extraArgument to reduxObservable

istarkov:
The same idea as at this [redux-thunk PR](https://github.com/gaearon/redux-thunk/pull/70)

(_People often request the ability to inject a custom argument into thunks._)

And I really need this ability ;-)

### DIFF
--- a/src/reduxObservable.js
+++ b/src/reduxObservable.js
@@ -2,14 +2,16 @@ import { Subject } from 'rxjs/Subject';
 import { from } from 'rxjs/observable/from';
 import { ActionsObservable } from './ActionsObservable';
 
-export function reduxObservable() {
+const DEFAULT_OPTIONS = {};
+
+export function reduxObservable({ extraArguments = [] } = DEFAULT_OPTIONS) {
   let actions = new Subject();
   let actionsObs = new ActionsObservable(actions);
 
   let middleware = (store) => (next) => {
     return (action) => {
       if (typeof action === 'function') {
-        let obs = from(action(actionsObs, store));
+        let obs = from(action(actionsObs, store, ...extraArguments));
         let sub = obs.subscribe(store.dispatch);
         return sub;
       } else {
--- a/test/reduxObservable-spec.js
+++ b/test/reduxObservable-spec.js
@@ -170,4 +170,24 @@ describe('reduxObservable', () => {
       done();
     }, 100);
   });
+
+  it('should pass options.extraArguments', (done) => {
+    const reducer = (state = [], action) => state.concat(action);
+
+    const options = { extraArguments: ['foo', 'bar'] };
+    const middleware = reduxObservable(options);
+
+    const store = createStore(reducer, applyMiddleware(middleware));
+
+    store.dispatch((_, __, foo, bar) => Observable.of({ type: 'ACTION', payload: `${foo}${bar}` }));
+
+    // HACKY: but should work until we use TestScheduler.
+    setTimeout(() => {
+      expect(store.getState()).to.deep.equal([
+        { type: '@@redux/INIT' },
+        { type: 'ACTION', payload: 'foobar' },
+      ]);
+      done();
+    }, 100);
+  });
 });


jayphelps:
In the future we'll likely have even more options to pass into the middleware creation process. So I'm thinking we need to either pass a hash of options, or use the same `withExtraArgument()` method redux-thunk does (though I think it should allow multiple args so people don't request that feature later.

```js
reduxObservable({ extraThunkArguments: [first, second, third] });
// or maybe
reduxObservable().withExtraArguments(first, second, third);
```

I'm leaning towards the first one, so future options we add are consistent, part of the options hash. Also, we'll need to have a test added for this new feature.

istarkov:
What the need of multiple arguments if we have in ES6 arguments destructuring assignment?

With current implementation I can pass any number of named arguments with simple syntax `reduxObservable({ cleverApiName, anotherApi }),`
and use them as 
`(actions$, store, { cleverApiName, anotherApi, ...etc }) => `

jayphelps:
Because not everyone uses ES2015 and mostly because it's trivial to make it work with multiple arguments so I don't see any downside? Let me know if you can think of some.

Did you have thoughts on my other suggestions about using a more generic options signature passed to `reduxObservable(options)` so that it's future-proof for adding additional options?

istarkov:
Having `options`, and the idea that `options` should be available at actions level, like
`(actions$, store, options) => ` is in't it is the same as PR above? IMHO there is no need to add  additional props like `extraThunkArguments` to public api. Just `options` or just `extraArgument` is enough for all ;-)

jayphelps:
sorry it wasn't clear. The options hash isn't for things passed to the thunks, it's literally for any and all middleware options. For example, we're considering giving people the option to choose a scheduler for testing purposes. I can see us adding other options as well. `extraThunkArguments` would be _just one of those options_. We would then take what's in `extraThunkArguments` and spread them onto the end of any thunk call.

istarkov:
Got it, you are right. What the method from both above u will prefer?

jayphelps:
I prefer `reduxObservable({ extraThunkArguments: [first, second, third] })` but I'm not married to it.

istarkov:
Changed.

blesh:
I think an options hash is fine. I have an issue up #25 that could use such a thing to pass a default scheduler for testing purposes.

blesh:
Overall, I think this PR is fine, I'd only like to bikeshed the name a little bit:

`extraThunkArguments` seems wonky to me... or at the least verbose. `addArgs` maybe? or `appendArgs`? 

I'm not sure. Thoughts?

istarkov:
May be like redux-thunk as projects somehow similar (both are middlewares ;-)) and name as `extraArguments`, or just `args`.

blesh:
I guess `extraArguments` is fine, as that's what `redux-thunk` seems to call it, internally.

istarkov:
Renamed.

blesh:
I wonder if it would be better to provide a "middleware" point... (lol, middleware-middleware). That would enable you to alter the observable before it's subscribed to.

So like:

```js
reduxObservable((obs) => obs.map(action => ({ ...action, ...additionalArgs })));
```

You could then do all sorts of crazy things. Delay all dispatched actions. Throttle them. I dunno.

@jayphelps?


Also, @istarkov, what is your use case? Or did I miss that above?

istarkov:
My usecase is really simple, just to inject external api, nothing more.

istarkov:
BTW in your case how u will setup a scheduler, etc?

blesh:
@istarkov digging into this and after a lot of talk with @jayphelps and @jhusain here at Netflix, we've pivoted back to more of our original design, which actually adds an argument to `reduxObservable`. (now published in 0.5.0) (beta software! #YOLO). So if we're going to add a config object, it would have to be the second argument to the middleware.

Also, looking at the use case... it seems like a standalone piece of middleware could solve that problem for you more elegantly. Something as simple as this for middleware would work:

```js
const addArgs = (...args) => (store) => (next) => (action) => next({ ...action, ...args })
```

blesh:
To be clear, @istarkov, I'm not saying "no" to this contribution. I'm asking your thoughts on using a separate middleware that serves the purpose of adding arguments to all actions.

Also, if this use case keeps coming up, we'll definitely be adding something for this.

istarkov:
Your `addArgs` is incompatible even with your `redux-observable` middleware, as `action` is a function in your case.

istarkov:
It will pollute every action in system, it will convert system based on plain js actions thunk midllewares to something there each action will have a functions as a props. It's not a widely used solution ;-)
At a first sight it is not what I want ;-)

istarkov:
There is a more simple idea to inject services in any middleware https://github.com/bradharms/redux-inject/blob/master/lib/index.js
BTW it need you to change every your action creator, not a good idea for a big projects.

blesh:
> It will pollute every action in system

But the proposed PR here will act on every action processed by the middleware, which is still very broad, no?

I think this is why solid use cases would be a good thing to gather to figure out what the best approach is for everyone.

For example, you could probably do something like:

```js
reduxObservable((actions) => actions.filter(a => a.requiresArgs).map(a => ({ ...a, extraArgs })))
```

And then you'd be able to dispatch actions you knew you wanted to add arguments to like so:

```js
dispatch({ type: 'ANYTHING', requiresArgs: true });

//or 

dispatch(() => Observable.of({ type: 'ANYTHING', requiresArgs: true }))
```

istarkov:
Now I understand the processor idea ;-) 

For now I does not use service like injections in my code (as in redux-thunk), I just use ApiMiddewares.
Redux dispatch has an ability to return values, I just call `dispatch({type: 'RETURN_API_I_NEED_NOW'})` and some of API middeware in the chain just return me the service I need.

But I start to hate all that midlewares zoo, all I need is something like your middleware and nothing more ;-) 

So you could close this PR as there are few solutions how to inject something in the chain, even I prefer current PR way I'm also not sure that Im 100% right, just looking on how `redux-thunk` changed I proposed this PR ;-)

istarkov:
For anyone who is interested in here is the source of any API middleware I use
```javascript
import { isFSA } from 'flux-standard-action';

export default function createApiMiddleware(api, actionType) {
  return function apiMiddleware(/* { dispatch, getState } */) {
    return next => action => {
      if (!isFSA(action)) {
        return next(action);
      }

      return action.type === actionType
        ? api(...action.payload)
        : next(action);
    };
  };
}
```

blesh:
@istarkov so the change made in 5.0 is satisfactory for your needs? I want to be sure you're happy or at least understand my responses to your PR.

Even more importantly, I want to make sure that I understand your concerns so I'm making the most informed decision I can. You seem like a good person, and I don't want to turn you away arbitrarily. I really appreciate your contribution.

istarkov:
Yes, I could use this library even without 5.0 change, and want to say that I really like this library approach.
As libraries author I understand that not every PR should be accepted ;-)  
If I can't find a solution or a best way to solve some problem the fastest way to solve it is to create a PR,  and even it will not be accepted I'll learn a lot and will get a solution.

Thank you for this library, for now it looks like it will solve all my needs in more simple way than sagas.

istarkov:
Thinking a little I get that `processor` is really what I need.
One question about `processor` as I see for now the only way to `unsubscribe` it, is to use `takeUntil` on processor stream?

blesh:
@istarkov that's correct. Honestly, using `takeUntil` to unsubscribe with RxJS _in general_ is more elegant in most cases. This is because calling `subscription.unsubscribe()` is very imperative, where using `takeUntil` is more more compositional and reactive.

istarkov:
Got it. Thank you.

istarkov:
The final result of my first redux-observable processor https://github.com/istarkov/redux-observable-process-fetch
It really allows to create a fun things:

Source:
https://github.com/istarkov/redux-observable-process-fetch/blob/master/src/createFetchProcess.js

Tests:
https://github.com/istarkov/redux-observable-process-fetch/blob/master/tests/createFetchProcess.spec.js

May be I'm the first who has published middleware for redux middleware ;-)

PR #31, opened 2016-05-20
feat(middleware processor): add argument to middleware to set up asyn…

blesh:
…c processing for all actions pumped through store

BREAKING CHANGE: dispatched actions will now occur _before_ the actions created by synchronous observable side effects.


blesh:
@jayphelps pushed that merge button, I swear.

PR #39, opened 2016-05-30
Update example comments

Sequoia:
"automatically dispatch" is misleading/confusing. This event is dispatched because `dispatch(fetchUser())` is called elsewhere-- merely returning an action object from `.map` function does not cause the action to be dispatched.


jayphelps:
Thanks! 👯

PR #40, opened 2016-06-03
fix(typings): `ofType` accepts multie keys

sebald:
Since 9027d1c the `ofType` method allows to specify multiple action types for comparison.
Adjusted TypeScript typings accordingly.


sebald:
Super mal patch :)

On a super unrelated, offtopic note: What made you guys write this middleware in ES6/7 rather than TypeScript? It is because of `::`?

PR #41, opened 2016-06-05
Simplify dispatchToProps in example

SimenB:
Wrapping in dispatch is what the second argument to dispatch does if passed an object, no need to do it manually


jayphelps:
Thanks @SimenB !

PR #43, opened 2016-06-07
Possible enhancements to the demo

cmelion:
I just want to share a few things that I found useful:

1) `npm install ` at project root failed for me, when running the basic demo, with a file not found error for redux-observable, I don't know why.  I ended up just doing `npm link`.

2) Added [JSON Server](https://github.com/typicode/json-server) to support quick mocking of ajax calls and altered demo to use it.

3) Added support for [Redux DevTools](https://github.com/gaearon/redux-devtools)

If any of these are useful, I can do individual PRs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,11 +18,7 @@ NOTE: This has a peer dependencies of `rxjs@5.0.*` and `redux`, which will have 
 as well.
 
 ```sh
-npm link
-
-cd examples/basic
-npm install
-npm start
+npm install --save redux-observable
 ```
 
 ## Usage
--- a/examples/basic/package.json
+++ b/examples/basic/package.json
@@ -20,7 +20,7 @@
     "react-dom": "^15.0.2",
     "react-redux": "^4.4.5",
     "redux": "^3.5.2",
-    "redux-observable": "^0.6.0",
+    "redux-observable": "*",
     "rxjs": "^5.0.0-beta.7"
   },
   "devDependencies": {
--- a/examples/basic/package.json
+++ b/examples/basic/package.json
@@ -4,7 +4,9 @@
   "description": "Example usage of redux-observable",
   "main": "server.js",
   "scripts": {
-    "start": "node server.js"
+    "server": "json-server --watch --delay 1000 json-server/db.json",
+    "client": "node server.js",
+    "start": "concurrently \"npm run server\" \"npm run client\""
   },
   "repository": {
     "type": "git",
@@ -18,7 +20,7 @@
     "react-dom": "^15.0.2",
     "react-redux": "^4.4.5",
     "redux": "^3.5.2",
-    "redux-observable": "*",
+    "redux-observable": "^0.6.0",
     "rxjs": "^5.0.0-beta.7"
   },
   "devDependencies": {
@@ -28,9 +30,12 @@
     "babel-plugin-transform-object-rest-spread": "^6.8.0",
     "babel-preset-es2015": "^6.6.0",
     "babel-preset-react": "^6.5.0",
+    "concurrently": "^2.1.0",
+    "json-server": "^0.8.14",
     "node-libs-browser": "^0.5.2",
     "react-hot-loader": "^1.2.7",
     "webpack": "^1.9.11",
-    "webpack-dev-server": "^1.9.0"
+    "webpack-dev-server": "^1.9.0",
+    "webpack-hot-middleware": "^2.10.0"
   }
 }
--- a/examples/basic/src/actions/index.js
+++ b/examples/basic/src/actions/index.js
@@ -6,7 +6,7 @@ export const FETCH_USER_ABORTED = 'FETCH_USER_ABORTED';
 const BASE_URL = 'http://localhost:3000/items/';
 
 export const fetchUser = () => (
-  (actions, store) => Rx.Observable.ajax(BASE_URL)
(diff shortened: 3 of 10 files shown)


README.md
```diff
@@ -18,7 +18,11 @@ NOTE: This has a peer dependencies of `rxjs@5.0.*` and `redux`, which will have
 as well.
 
 ```sh
-npm install --save redux-observable
```
jayphelps: This is instructions on how to install redux-observable into your own project, not how to install the and run the examples.

examples/basic/src/actions/index.js
```diff
@@ -3,10 +3,11 @@ import * as Rx from 'rxjs';
...
 export const FETCH_USER_FULFILLED = 'FETCH_USER_FULFILLED';
 export const FETCH_USER_ABORTED = 'FETCH_USER_ABORTED';
+const BASE_URL = 'http://localhost:3000/items/';
 
 export const fetchUser = () => (
-  (actions, store) => Rx.Observable.of({ id: 1, name: 'Bilbo Baggins', timestamp: new Date() })
-    // Delaying to emulate an async request, like Rx.Observable.ajax('/api/path')
+  (actions, store) => Rx.Observable.ajax(BASE_URL)
+    // Delaying to provide enough time to cancel manually
     .delay(1000)
```
jayphelps: If we're already delaying ajax requests on the json-server (w/ `--delay 1000`, I think it's best to remove this so it's more "real world".

README.md
jayphelps: If instead you want to clarify the process better, the full instructions could be amended to the `examples/basic/README.md` to be:

```
git clone https://github.com/redux-observable/redux-observable.git
cd redux-observable
npm install
cd examples/basic/
npm install
npm start
```

examples/basic/package.json
```diff
@@ -18,7 +20,7 @@
     "react-dom": "^15.0.2",
     "react-redux": "^4.4.5",
     "redux": "^3.5.2",
-    "redux-observable": "*",
+    "redux-observable": "^0.6.0",
```
jayphelps: Don't want to hardcode the version number because then every time we bump it we have to update it here too. What's listed here really doesn't matter because we alias `redux-observable` to source code in the project root here: https://github.com/redux-observable/redux-observable/blob/master/examples/basic/webpack.config.js#L25

jayphelps:
This is looking good. 👍 ping me if you have thoughts on my comments or when you've made the suggested changes. Thanks dudebro!

examples/basic/src/actions/index.js
jayphelps: Does this still work as expected? I can't test it at the moment but wouldn't this not parse the JSON? `Rx.Observable.ajax.getJSON` is a helper to do that, if it does have that problem.

examples/basic/src/actions/index.js
```diff
@@ -3,11 +3,10 @@ import * as Rx from 'rxjs';
 export const FETCH_USER_PENDING = 'FETCH_USER_PENDING';
 export const FETCH_USER_FULFILLED = 'FETCH_USER_FULFILLED';
 export const FETCH_USER_ABORTED = 'FETCH_USER_ABORTED';
+const BASE_URL = 'http://localhost:3000/items/';
 
 export const fetchUser = () => (
-  (actions, store) => Rx.Observable.of({ id: 1, name: 'Bilbo Baggins', timestamp: new Date() })
-    // Delaying to emulate an async request, like Rx.Observable.ajax('/api/path')
-    .delay(1000)
+  (actions, store) => Rx.Observable.ajax(BASE_URL)
```
jayphelps: Does this still work as expected? I can't test it at the moment but wouldn't this not parse the JSON? `Rx.Observable.ajax.getJSON(url)` is a helper to do that, if it does have that problem.

cmelion:
@jayphelps makes sense.  Changes made.

examples/basic/src/actions/index.js
cmelion: @jayphelps, you are correct, this needs more work to display the expected data.
The reducer should also set the timestamp instead of hardcoding in the data.

jayphelps: @cmelion cool. possible for you to make those changes? 💃

cmelion: @jayphelps  I think that covers it.  Thanks!

jayphelps: Cool. Will test locally this afternoon to confirm it works as expected and then merge. 👍

cmelion: ahh wait, the timestamp is not fixed correctly!  Will fix shortly.

jayphelps:
Merged as 3568368522d51b0d114b6bec21f54941b905109c

Thanks again!

examples/basic/src/reducers/index.js
```diff
@@ -7,7 +7,7 @@ const user = (state = { isLoading: false }, action) => {
       return { ...state, isLoading: true };
 
     case FETCH_USER_FULFILLED:
-      return { ...state, isLoading: false, ...action.payload };
+      return { ...state, isLoading: false, ...action.payload, timestamp: new Date() };
```
cmelion: Calling new Date() here violates the convention that reducers should be pure functions.

Issue #47, opened 2016-06-21
Question re: SSR


StevenLangbroek:
If I use React Router's `match` function, and trigger actions from all matched components, is there a good way to coordinate when all these actions (Observables) are `completed`? This would be essential for SSR, and is normally coordinated with Promises and `Promise.all`.

istarkov:
You could add a redux-observable processor with server process like 

https://jsfiddle.net/u1gay0qj/

```javascript
Rx.Observable.combineLatest(
	// use actions$.ofType('MY_ACTION')
  actions$.filter(a => a.type === 'MY_ACTION').take(1),
  actions$.filter(a => a.type === 'MY_OTHER_ACTION').take(1),
  // all other actions
)
.delay(0)
.do(() => console.log('HERE YOU CAN SAVE STATE'))
```

istarkov:
In my case as I use for data fetch standalone process, like this https://github.com/istarkov/redux-observable-process-fetch/blob/master/src/createFetchProcess.js
I just could count `LOADING_START`, `LOADING_END` (`_CANCEL`) actions, and when 
`count(LOADING_START) === count(LOADING_END)` I could save state

jayphelps:
@StevenLangbroek To be honest, neither myself or @blesh have much experience with React SSR since none of the apps we work on need it.

This is a great question that we _should_ provide guidance on..we will either need to familiarize ourselves with React SSR or defer to the experiences of some of our community members. As @istarkov mentioned, we recently introduced the ability to define "process managers" that are very similar conceptually to redux-saga (which are actually process managers too, not _true_ sagas since they maintain state) so I imagine there will be existing solutions we can learn from. e.g. https://github.com/yelouafi/redux-saga/blob/master/examples/real-world/server.js but you'll quickly see that redux-saga is not redux-observable--for one, there's no first-class concept of tasks.

We're working on a major overhaul of docs, I'll add this to the list of items to explore. If anyone has ideas, feel free to share them here.

StevenLangbroek:
@jayphelps thanks for the reply. Conceptually speaking, the process is relatively straight forward:

- match routes to an array of components
- pick static methods off of them
- map over them and call them with `dispatch`, `params`, `query` and what else you need to be able dispatch actions for the routes
- coordinate their total completion somehow, and carry on with rendering

I'm less experienced with rx / observables, so the 2 main questions I'd have to be able to chime in is:

[...]

PR #48, opened 2016-06-21
npmcdn support

istarkov:
To avoid examples [like in this comment](https://github.com/redux-observable/redux-observable/issues/47#issuecomment-227593746) (sorry ;-))

Let's just publish redux-observable on [npmcdn](https://npmcdn.com)
All this project needs is umd versions at `dist` folder in npm.

So it will be avalable after `npm version patch; npm publish` at 
`https://npmcdn.com/redux-observable@latest/build/redux-observable.js`
`https://npmcdn.com/redux-observable@latest/build/redux-observable.min.js`

And can be accessed via global `ReduxObservable` var. At `jsbin` and similar playgrounds.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -251,6 +251,7 @@
   },
   "env": {
     "browser": true,
+    "node": true,
     "es6": true
   },
   "globals": {
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules
 lib
 temp
+dist
--- a/package.json
+++ b/package.json
@@ -5,12 +5,15 @@
   "main": "lib/index.js",
   "scripts": {
     "lint": "eslint src && eslint test",
-    "build": "npm run lint && rm -rf lib && babel src -d lib",
+    "build": "npm run build:commonjs && npm run build:umd && npm run build:umd:min",
+    "build:commonjs": "babel src -d lib",
+    "build:umd": "cross-env NODE_ENV=development webpack src/index.js dist/redux-observable.js",
+    "build:umd:min": "cross-env NODE_ENV=production webpack src/index.js dist/redux-observable.min.js",
     "build_tests": "rm -rf temp && babel test -d temp",
-    "clean": "rimraf ./lib; rimraf ./temp;",
+    "clean": "rimraf lib temp dist",
     "test": "npm run build && npm run build_tests && mocha temp && npm run test_typings",
     "test_typings": "tsc test/typings.ts --outFile temp/typings.js --target ES2015 --moduleResolution node",
-    "prepublish": "npm test"
+    "prepublish": "npm run clean && npm run lint && npm test"
   },
   "typings": "./index.d.ts",
   "files": [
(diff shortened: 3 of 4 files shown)


jayphelps:
Beautiful.

Though for my jsbin example, I really did have to re-implement most of reduxObservable to support what I was demoing, so wouldn't have matters 😋

jayphelps:
@istarkov Although we don't _currently_ import `redux` anywhere it's prolly a good idea to go ahead and include `externals` so if we ever do, we don't start accidentally packaging it along for the ride.

Which makes me wonder...should we bundle `rxjs` as this PR currently does or also `externals` it? I'm inclined to `externals` it as well, assuming it works with the RxJS UMD build..

jayphelps:
@istarkov btw this is what I mean by `externals` https://github.com/jayphelps/redux-transaction/blob/master/webpack.config.babel.js#L38-L45

istarkov:
I'll check now

istarkov:
Even this works 

```
  externals: {
    'rxjs/Observable': 'Rx',
    'rxjs/Subject': 'Rx',
```

But  there is a problem with `externals`, with Rx operators

```
    'rxjs/observable/from': ?
    'rxjs/operator/filter': ?
    'rxjs/observable/merge': ?
```

istarkov:
This problem will gone if to rewrite code on operators like `import 'rxjs/add/observable/from';`

jayphelps:
The problem with using that form is that then weird things can happen to users in future releases if they accidentally forget to import the same operator themselves.

e.g. if we `import 'rxjs/add/observable/from';` right now, and devs consuming our lib need to use `from` too but _forget_ to import it, it will probably work right now assuming they import redux-observable first. But if we push another release which no longer needs to do `import 'rxjs/add/observable/from';`, then their code breaks because it was only working because of us importing it.

This might seem unlikely, but I've actually already been bitten by it a couple times with libs we maintain internally. I'm not 100% "no" to it...just would prefer to dig a big longer.

istarkov:
Solved, looks like webpack has a lot of undocumented (or I just not found) features

Tested with
```
<html>
  <head>
  <script src="https://npmcdn.com/@reactivex/rxjs@5.0.0-beta.9/dist/global/Rx.umd.js"></script>
  <script src="./dist/redux-observable.js"></script>
  <script>
    let actions = new Rx.Subject();
    let actionsObs = new ReduxObservable.ActionsObservable(actions);
    const processor = ReduxObservable
      .combineDelegators((actions$) => actions$.ofType('X')
      .do(() => console.log('X'))
      .filter(() => false));
    processor(actionsObs, {}).subscribe(x => console.log(x));
    actions.next({ type: 'X' });

  </script>
  </head>
  <body>
  </body>
</html>
```

istarkov:
image
Minified build is just 3.33Kb

jayphelps:
Nice! I think this is the best we're gonna do at the moment. My slight concern now is every time we use a new operator we need to remember to add it to the `externals`..the only way around this seems to be to generate the full list of possible rxjs externals...but that's non-trivial soooo....punting on that one!!

We'll just have to be careful and periodically audit the umd build to be sure we're not including operators that should be externed.

@blesh c/d? this LGTM

jayphelps:
Turns out it's not _that_ hard to generate the externs on the fly. I PR'd that ability to your PR. https://github.com/istarkov/redux-observable/pull/1

We'll hold off on merging this until you either merge my PR to yours or rebut.

istarkov:
Merged.

jayphelps:
Thanks @istarkov!! we'll publish a release soon.

Issue #38, opened 2016-05-27
HELP: what do we name this new feature?


jayphelps:
Although undocumented at the moment, we've added the ability to compose and inject *static* action handlers directly into the middleware. That is, they are not dispatched, instead you dispatch redux action POJOs and react to those actions. This keeps your actions always pure objects and allows you to do debouncing and action composition easier in many ways.

Here are two examples, fetching a user by id (as a trivial example) and a fairly non-trivial search example that debounces value changes, cancels in-flight requests, server error response handling, and even supports pagination. 

```js
// FETCH USER

const fetchUser = id => ({ type: FETCH_USER, payload: id });

const fetchUserDelegator = actions =>
  actions.ofType(FETCH_USER)
    .switchMap(action =>
      Observable.ajax.getJSON(`/api/users/${action.payload}`)
        .map(payload => ({ type: FETCH_USER_FULFILLED, payload }))
        .startWith({ type: FETCH_USER_PENDING })
    );

// SEARCH

const setPage = value => ({ type: SET_PAGE, value });
const searchValueChanged = value => ({ type: SEARCH_VALUE_CHANGED, value });
const searchFulfilled = value => ({ type: SEARCH_RESULTS_FULFILLED, value });
const searchErrored = value => ({ type: SEARCH_RESULTS_ERRORED, value });
const searchPending = () => ({ type: SEARCH_RESULTS_PENDING });

const searchDelegator = (actions, { getState }) =>
  Observable.combineLatest(
    actions.ofType(SEARCH_VALUE_CHANGED)
      .debounceTime(250)
      .map(action => action.value)
      .startWith(getState().search.value)
      .filter(value => !!value),
    actions.ofType(SET_PAGE)
      .map(action => action.value)
      .startWith(getState().search.page),
    (q, page) => `https://api.github.com/search/repositories?q=${q}&page=${page}&per_page=${resultsPerPage}`
  )
    .switchMap(url =>
      Observable.ajax.getJSON(url)
        .map(searchFulfilled)
        .catch(({ xhr }) => Observable.of(searchErrored(xhr.response.message)))
        .startWith(searchPending())
    );

// STORE SETUP

const rootDelegator = combineDelegators(
  fetchUserDelegator,
  searchDelegator
);

const store = createStore(rootReducer,
  applyMiddleware(
    reduxObservable(rootDelegator)
  )
);
```

[...]

PR #50, opened 2016-06-22
update typings

istarkov:


jayphelps:
regarding `Processor`, in #38 we decided to use the term "process manager" to describe these...but while that's an apt description it's a very verbose. I'm not thrilled about things like `combineProcessManagers`. Perhaps we document them as "process managers" but use a shortened term for them in code and common usage...something like `Manager`

ugh...naming things... haha

istarkov:
Naming is not an easy task ;-) 
Looks like it will be better to wait documentation and source changes. (Where Managers could become a Partners or anything like ;-))
Also I see now that this PR is incomplete and could be closed or used as like a note, that current typings is outdated. 
(I'm not using typescript in my daily work, so can't be sure that it's good in all other cases, it will be better if some real ts user provide this)

jayphelps:
@istarkov agree with all your points. We should wait until docs are settled. This is number 1 on my OSS priority for my vacation next two weeks.

Issue #52, opened 2016-06-24
Add ability to add new delegator


alisd23:
In an app which uses code splitting, reducers might be loaded lazily as the user navigates. It would nice to be able to add delegators to the root delegator after the app is run initially to support these kinds of applications.

istarkov:
You could just `dispatch` loaded delegator, and it will work the same as processor delegators.
```
dispatch((action$, store) => action$.blabla(....));
```
See the source https://github.com/redux-observable/redux-observable/blob/master/src/reduxObservable.js#L14

martypenner:
That's really only for thunkservables, though. The processor [here](https://github.com/redux-observable/redux-observable/blob/master/src/reduxObservable.js#L10) is the root delegator/process manager.

istarkov:
@martypenner for now there is just one difference from
[this](https://github.com/redux-observable/redux-observable/blob/master/src/reduxObservable.js#L11)
```
processor(actionsObs, store).subscribe(store.dispatch);
```
and [this](https://github.com/redux-observable/redux-observable/blob/master/src/reduxObservable.js#L15-L17)
```
let obs = from(action(actionsObs, store));
let sub = obs.subscribe(store.dispatch);
return sub;
```

you could unsubscribe from second.
Both will work in exactly the same manner. So having that your delegators are loadable, you can use `dispatch` variant

istarkov:
In the future having [this comment](https://github.com/redux-observable/redux-observable/issues/44#issue-160238580) `I, personally, would like to deprecate the thunkservable method prior to RC, but I feel like @jayphelps might have different opinions.` this method could become a problem, but now it work.

jayphelps:
(delegators == old term for what we're currently calling "process managers")

Because process managers just return a stream, you can already implement this yourself--though we do intend to support something like `replaceManager(newRootManager)` which would just do the same thing you could do under the hood.

When I get a sec I'll post the code to do this in the meantime.

istarkov:
```javascript
const dynamicD$ = new Rx.Subject();
const combineDelegators = (dynamicDelegators$, ...delegators) => (actions, store) =>
    dynamicDelegators.mergeMap(dynamicD => dynamicD(actions, store))
    	.merge(...(delegators.map((delegator) => delegator(actions, store))));

// .....
// to add call
dynamicD$.next((action$, store) => blabla)
```

alisd23:
[...]

PR #53, opened 2016-06-24
Dynamically add epics after store creation

alisd23:
For issue #52

### DIFF
--- a/src/combineDelegators.js
+++ b/src/combineDelegators.js
@@ -1,7 +1,18 @@
 import { merge } from 'rxjs/observable/merge';
+import { Subject } from 'rxjs/subject';
 
 /**
   merges all delegators into a single delegator.
  */
-export const combineDelegators = (...delegators) => (actions, store) =>
-    merge(...(delegators.map((delegator) => delegator(actions, store))));
+export const combineDelegators = (...delegators) => {
+  const dynamicDelegators$ = new Subject();
+
+  const delegator = (actions, store) =>
+    dynamicDelegators$
+      .mergeMap(dynamicD => dynamicD(actions, store))
+      .merge(...(delegators.map((delegator) => delegator(actions, store))));
+
+  delegator.addDelegator = (newDelegator) => dynamicDelegators$.next(newDelegator);
+
+  return delegator;
+}
--- a/src/combineDelegators.js
+++ b/src/combineDelegators.js
@@ -1,4 +1,4 @@
-import { Subject } from 'rxjs/subject';
+import { Subject } from 'rxjs/Subject';
 
 /**
   merges all delegators into a single delegator.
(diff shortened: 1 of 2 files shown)


jayphelps:
We'll have to sit on this for a bit cause I imagine we want to instead follow redux's pattern of having a `replaceDelegator(rootDelegator)` analogous to `replaceReducer(rootReducer)`. Then if we have that is some ways it makes this redundant--though I'm not writing it off completely cause it does in fact reduce boilerplate even with `replaceDelegator`.

alisd23:
No problem, I'm interested to see a **replaceDelegator** solution

jeffbski:
`replaceDelegator` would be nice, but if that is too difficult to solve due to accumulated state then `addDelegator` would solve the need for lazy loading and I believe would be significantly easier to implement. So we could consider that as a solution which meets 90% of common needs.

jeffbski:
And I didn't realize that this PR actually had implemented addDelegator already. So it is already done and quite simple. If replaceDelegator is on the long term plan, then I'd vote for getting this in for now and it can be removed if and when replaceDelegator is ready in the future.

jayphelps:
Our current thinking is instead of providing the ability to add them (cause then we'd need a way to remove them) we'll just provide a way to replace it. Works for both this case and the hot reloading case too and is more inline with redux's `replaceReducer`.

Check out https://github.com/redux-observable/redux-observable/pull/75

jayphelps:
Here's simple Rx for anyone who wants to add new Epics without tearing down the old ones:

http://jsbin.com/tuxikuk/edit?js,output

```js
const epic$ = new BehaviorSubject(combineEpics(epic1, epic2));
const rootEpic = (action$, store) =>
  epic$.mergeMap(epic =>
    epic(action$, store)
  );

// sometime later...add another Epic, keeping the state of the old ones...
epic$.next(asyncEpic1);
// and again later add another...
epic$.next(asyncEpic2);
```

jayphelps:
Hey @alisd23, super appreciate your work on this. We did decide to keep the middleware only supporting a single, root epic for now. Hope you understand 👯 

As with most things, we'll continue to evaluate this so it may one day make sense to add this functionality. For now, people can fairly trivially brew their own method, like:

```js
const epic$ = new BehaviorSubject(combineEpics(epic1, epic2));
const rootEpic = (action$, store) =>
  epic$.mergeMap(epic =>
    epic(action$, store)
  );

// sometime later...add another Epic, keeping the state of the old ones...
epic$.next(asyncEpic1);
// and again later add another...
epic$.next(asyncEpic2);
```

Thanks again!

alisd23:
@jayphelps Sorry totally forgot about this...
No problem :+1:

asvishnyakov:
https://github.com/redux-observable/redux-observable/issues/525#issuecomment-397431316

PR #55, opened 2016-06-27
WIP(docs): Add initial documentation using gitbook

### LINKED ISSUE #44: Document processManager
Edit: WIP docs are up, please review! https://github.com/redux-observable/redux-observable/pull/55

I strongly feel that the best and most powerful way to use redux-observable is with processManagers rather than thunkservables. Unfortunately, right now we're only documenting the thunkservable methodology.

I, personally, would like to deprecate the thunkservable method prior to RC, but I feel like @jayphelps might have different opinions.

Either way, we really need to document it in a first class way.

jayphelps:
These docs call the process manager style "Actions Managers" but this isn't a great name. Please vote on a new name: http://www.poll-maker.com/poll739209x907c4240-30
--
Not entirely ready yet but wanted to open the floor for comments and other feedback.

To view the full HTML website for the docs, use `npm run docs:watch`, or view them as [markdown here](https://github.com/redux-observable/redux-observable/tree/docs/docs)

We need to update the [runnable examples](https://github.com/redux-observable/redux-observable/tree/docs/examples/basic) too (and add a lot more of them) but I'm thinking that should be a separate effort so we don't block this doc update for too long.

I made several very opinionated assumptions in here, so please feel free to call me out on them. Some of them are the renaming of `combineDelegators` to `combineManagers` but the biggest is the use of the term "Actions Manager" instead of just "Process Manager" which, while writing the docs, I felt was potentially confusing since most people associate the word "process" with OS-level processes and not the general CS term.

I'm thinking we'll utilize https://js.org/'s DNS service so docs will be at http://redux-observable.js.org/docs/, hosted on Github Pages.

Will eventually close #44 

### TODO (before merge)
- [x] setup some sort of private email for CoD/security concerns that forwards to team members
- [x] favicon
- [ ] Update top-level README
- [ ] Actually deprecate thunkservables
- [ ] Change actual API of `combineDelegators` to `combineManagers`
- [ ] Verify all code examples actually work
- [x] note that actions are received _after_ reducers have already received it.

### TODO (deferred until some time after merge)
- [ ] consider providing React, Angular, Ember usage in the docs, just for complete clarity of a start to finish example


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,74 @@
...
+when an individual is representing the project or its community. Examples of
+representing a project or community include using an official project e-mail
+address, posting via an official social media account, or acting as an appointed
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting the project team at dan.abramov@me.com. All
```
istarkov: `dan.abramov@me.com` sure?

jayphelps: Nope! Obviously I boosted this from redux and didn't catch that. Thanks for catching!!!! Really appreciate you're reading things thoroughly.

docs/api/README.md
```diff
@@ -0,0 +1,2 @@
+## API Reference
+
```
jayphelps: TODO: missing links to the sub docs and need the middleware API docs

docs/FAQ.md
```diff
@@ -0,0 +1,33 @@
...
+  - [Why were thunkservables deprecated?](#general-thunkservables-deprecated)
+
+## General
+
+<a id="general-rxjs-v4"></a>
+### Does this work with RxJS 4/most.js/bacon.js? 
+
+##### RxJS v4
+
+Not out of box. We'd love to provide an elegant for solution for this soon, ideally [RxJS v4 would add interop with `Symbol.observable`](https://github.com/Reactive-Extensions/RxJS/issues/1225).
```
istarkov: RxJs 4 support could be done as like in recompose https://github.com/acdlite/recompose/blob/master/src/packages/recompose/rxjs4ObservableConfig.js

docs/basics/ActionsManagers.md
```diff
@@ -0,0 +1,194 @@
...
+> Noticed how `action$` has a dollar sign at the end? It's simply a common RxJS convention to identify variables that reference a stream.
+ 
+`pingManager` will listen for actions of type `PING` and map them to a new action, `PONG`. This example is functionally equivalent to doing this:
+
+```js
+dispatch({ type: 'PING' });
+dispatch({ type: 'PONG' });
+```
+
+> Actions Managers run alongside the normal Redux dispatch channel, so you cannot "swallow" an action. When you map one action to another, you're not preventing the original action from reaching the reducers.
```
jayphelps: This needs to be more clear that it's referring to incoming actions, not outgoing ones.

docs/basics/ActionsManagers.md
```diff
@@ -0,0 +1,194 @@
...
+  action$.ofType(INCREMENT_IF_ODD)
+    .filter(() => store.getState().counter % 2 === 0)
+    .map(increment);
+);
+
+// later...
+dispatch(incrementIfOdd());
+```
+
+Remember, `store.getState()` is just an imperative, synchronous API. You cannot treat it as a stream as-is. While it's not a common pattern for Managers to need, Redux does support a stream of state changes:
```
jayphelps: I'd like to propose we consider providing a `state$` argument as a stream (still keeping the store provided too as escape hatch). Will file a ticket with my reasoning.

docs/api/combineManagers.md
```diff
@@ -0,0 +1,26 @@
+# `combineManagers(...managers)`
```
istarkov: ;-) http://stackoverflow.com/questions/1866794/naming-classes-how-to-avoid-calling-everything-a-whatevermanager

jayphelps: Yeah. We've bikeshed a ton (mostly in person) and we just don't have a catchy name. Process manager was the most apt definition but is a horrible name. i really wish we could come up with a clever name that has no real meaning in tech previously, like "saga" was (before someone coined it ". I DUNNO........ 😖

docs/basics/ActionsManagers.md
```diff
@@ -0,0 +1,194 @@
...
+
+You can think of it of having roughly this type signature:
+
+```js
+function (action$: Observable<Action>, store: Store): Observable<Action>;
+```
+
+While you'll most commonly produce actions out in response to some action you received in, keep in mind that's not actually a requirement! Once you're inside your Manager, use any Observable patterns you desire as long as anything output from the outermost stream is an action.
+
+The actions you emit will be immediately dispatched through the normal `store.dispatch()`, so under the hood redux-observable effectively does `manager(actions$, store).subscribe(store.dispatch)`
```
istarkov: IMO here should be written that not filtering initial action, could cause infinite loop, i.e
`const actionManager = (action$) => action$` will cause infinite loop

docs/api/combineManagers.md
istarkov: May be call them `Daemons` (unix daemon analogy) as it is somehow a background process, not interactive, and can respond (but not change) certain actions (system events).

alisd23: I quite liked the `coordinator` suggestion. The only problem is that it's quite a long word for variable names

ctrlplusb: Thanks @alisd23 - agreed it's quite a long one. 

I really like the `daemon` suggestion by @istarkov.  I was actually on my way here to suggest an alternative of `epic`, but the `daemon` suggestion seems more intuitive.

docs/basics/ActionsManagers.md
blesh: That would just be sugar for `Observable.from(store)`. I'm indifferent either way. We can either provide guidance to do so, or we can just provide the argument. The only thing I'd want to see before adding it to the API would be some half-reasonable argument for the demand to use it. Doesn't even need to be a good argument, since it's a trivial add.

blesh:
@jayphelps most of this looks great. Since documentation is important and will evolve, is it possible to just merge this and evolve it in place, rather than "big bang" it in. I feel like docs are a little different than tech features.

docs/basics/ActionsManagers.md
jayphelps: 👍 Indeed. Let's move discussion to #56

docs/FAQ.md
jayphelps: @istarkov can you create a ticket with your full thoughts on how this would work?

docs/api/combineManagers.md
jayphelps: Actually, we did consider coordinator , daemon, and epic...I really liked epic since it's similar in meaning to saga but still fairly novel in CS (besides the agile term)..but we always had (relatively weak) excuses for not choosing them. I'm at the point where I just want almost _anything_ as long as the community mostly doesn't hate it. At least in this thread, it sounds like daemon is liked by many. Any opinions @blesh ?

docs/basics/ActionsManagers.md
jayphelps: Excellent suggestion!!

jayphelps: Added https://github.com/redux-observable/redux-observable/pull/55/files#diff-79179b01416603378e91e2e0e12d8a6bR22

docs/api/combineManagers.md
margaretmoser: I like `daemon` too. Did you also consider `pipe` (takes a stream, returns a stream)?

jayphelps: @margaretmoser considered it, but it's not actually a pipe. The output of one is _not_ the input of another. Instead they all share the same input and their outputs are merged together.

istarkov: I know the term but it's not cool - SPMC (single producer multiple consumers) ;-) from the world of lock free algorithms

margaretmoser: I doubt I could come up with anything you haven't already considered. Some generic terms I've seen in descriptions of message-passing systems are `channel`, `messagebroker` or just `broker`, `center` (as in `NSNotificationCenter`), or just the good ol' `publisher`. Not sure I like any of those better than `daemon`, personally.
`</bikeshed>`

jayphelps: @margaretmoser we hadn't considered channel, broker, center so keep them coming! I may have accidentally implied that repeating ones wasn't productive but it actually is. Others may rally around ones we initially dismissed, which we did mostly because we felt the community wouldn't like it.

matthewwithanm: FWIW the concept reminds me of [workers](https://developer.mozilla.org/en-US/docs/Web/API/Worker).

martypenner:
I'm currently using redux-observable in combination with [redux-optimist](https://github.com/ForbesLindesay/redux-optimist). Some actions I'm dispatching are incompatible with redux-optimist's way of doing things, so I've created a thunkservable that covers the gaps. I need to emit a few actions BEFORE the reducers get hit, then emit a few after as well. It would be nice to have some mechanism in the process managers that allowed this.

docs/api/combineManagers.md
margaretmoser: Well then, some more random thoughts:
`Aggregator` (as in [event management](http://www.donnfelker.com/wp-content/uploads/WindowsLiveWriter/UsingPrismsEventAggregator_70CA/EventAggregator_3.png))
`Invoker` (as in Command Pattern)
`Container` (based on @matthewwithanm's suggestion, see [ServiceWorker](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API))

One other thought: could you change it to a noun, something like `CombinedAction` or `ActionStream`, or, I dunno, `MulticastAction`? I mean, it seems like more of an extension/decoration of Actions to me than a separate object that manages them. But I am still finding my way around, could be way off base.

blesh:
Epic, daemon... but consider this:

> A group of ducks can be called a flock, brace, raft, team or paddling. A group of ducks is referred to as a flock while they are in flight. They are more often referred to as a raft, team or paddling while the group is on water.

:D  Yassssss

jayphelps:
@martypenner https://github.com/redux-observable/redux-observable/pull/55#issuecomment-229109967 can you create a new ticket and in it share the details of your use case a bit further like what things you need to do _before_ reducers are hit in the case of redux-optimist?

docs/api/combineManagers.md
ascarius: And why not `Effect` like [ngrx/effects](https://github.com/ngrx/effects) ?

martypenner:
@jayphelps I've add [an issue here](https://github.com/redux-observable/redux-observable/issues/59).

docs/api/combineManagers.md
jayphelps: OK peeps, I made a poll. Please vote on a new name or present a better one! 😄  http://goo.gl/vNULnb

jayphelps: @matthewwithanm IMO that would be particularly confusing and potentially confusing, make people think they naturally run in workers for performance or something. But I fee ya!

r-park: **Redux Effects** and **Redux Epics** sound best to me.

jayphelps: @r-park make sure you vote 👍 http://goo.gl/vNULnb

r-park: Voted, but fyi it looks like the poll allows you to vote multiple times — was that intentional?

jayphelps: @r-park shhhhh 😋 we're on the honor system here. Let's all be adults.

ascarius: It seems we have a winner :+1:
I'm glad of this result. It would have been pretty awkward than `@ngrx` and `redux-observable` use a different name.
I'm very impatient to use it :hatched_chick:

blesh:
As @jayphelps have been working with this at Netflix, we've become partial to the name "Epic". This is for a few reasons:

1. An epic can contain many sagas. This is true for the functionality here too. One can add many separate sagas to the logic in this area.
2. The term "epic" is fairly terse.
3. The agile term "epic" is unrelated enough to the functionality, that we have limited concerns about that clashing.

We understand that the vote had "effect" winning by votes. But the sample size was small and the margin in that situation is questionable. There are also a few things that we don't like about "effect".

1. This functionality will almost always contain more than one side-effect. Therefor it should probably be "effects" if anything, but pluralizing the name doesn't make sense.
2. The noun itself feels clunky.  "mainEffect" being a collection of other effects? "loadEffect", "saveEffect"... each of those likely having a couple of side effects?

The entire purpose of this functionality is to:

1. observe one or more specific sets of actions
2. initiate side effects and async logic from those actions
3. dispatch more actions based off of the results and/or errors from those side effects and async calls.

Given the above, I don't think "effect" fits, where I do think "epic" fits, at least a little better.

jayphelps:
We realize this is dragging out pretty long..we really _really_ don't want to have to change the name yet-again so we're bikeshedding the crap out of it so we can hopefully not do this ever again. We do care a lot about the communities opinion, we just felt we didn't make a clear case for epic (and why we didn't like effect) so we're taking a sort of "last stand". If you feel strongly, please do voice your thoughts!

matthewwithanm:
I'm not really fond of "meaningless" names—especially one chosen for its relation to another library's meaningless name and especially when there seem to be so many analogs already (workers, greenlets, etc). But whatever you call it, I wanna use it 😝

jayphelps:
@matthewwithanm I feel ya, but feel pretty strongly that this is novel enough to justify it, just like "saga" was novel enough at the time--though not redux-saga's use of "saga", which isn't true to the definition.

I believe we're open to arguments still, if you think there is an existing name that truly does describe it in a non-confusing way. ngrx/effects isn't really a widely used library so I don't currently think it bears much weight.

istarkov:
Want to say that for me with non native english epic does not look cool, as 
the only translation of epic to russian exists and this word uses in my language as `epic battle` == `heroic battle`. What means epic in your case?

I also think, that whatever you call it, I wanna use it.

jayphelps:
@istarkov 

> ep·ic (noun)
> 1. a long poem, typically one derived from ancient oral tradition, narrating the deeds and adventures of heroic or legendary figures or the history of a nation.

joshburgess:
I agree that "effects" is too generic... there is already a redux-effects library and also redux-loop, which is a port of elm-effects.

Although not the same idea, this conversation is making me think of concepts like **actors** and **supervisors** in Erlang/Elixir/Akka/etc... I'm not sure that's at all helpful, but maybe it will spark some creativity in someone. "Supervisors" **sort of** makes sense...

And although the name "coordinators" isn't too catchy, definitions for coordinate definitely describe the functionality elegantly...

> 
> co·or·di·nate
> verb
> kōˈôrdəˌnāt/
> - _bring the different elements of (a complex activity or organization) into a relationship that will ensure efficiency or harmony._
> - _negotiate with others in order to work together effectively._
>

Other ideas: arbiters, brokers, directors, moderators, negotiators 

I don't know... I still like processes, process managers, supervisors, etc. as they are relatively common terms and make sense. I mean, action dispatching is really just message passing, and that just makes me think of Erlang/Elixir's communicating processes... Although, because the concept is built with streams here, it's sort of its own thing in this case, I suppose.

ascarius:
Ha ha! You guys are using democracy like a french government and THAT is EPIC! (sorry for this bad joke)

IMO, Epic make sense if we think about sagas, but the big problem is this is an adjective!

What about "Reaction" ?
Because action -> reaction.
And because reactions can be chained!

blesh:
> What about "Reaction" ?

@ascarius "I'm using Reactive Extensions with React and React-Observable to create Reactions so I can use Reactive Programming to manage async"

😂

matthewwithanm:
Job?

matthewwithanm:
Job?

**EDIT:** Eh, it's more of a job runner than a job. "Runner" might work. I keep coming back to how similar this to workers/goroutines/threadlet/etc. I'm trying to think of something that plays off that—something that says "like that but with observables!—but coming up blank. Reaction does seem appropriate if "react" weren't so overused. "Consequence" plays off the same cause/effect action/reaction idea ("Actions have consequences").

jayphelps:
I totally borked the git history on this PR and github refuses to reopen it after I fixed it, so had to move it into a new one at #66. It's ready, using the name "Epic". 

We need to land #67 first, then will merge docs, shortly there after will publish them up on http://redux-observable.js.org

Issue #13, opened 2016-05-14
Compatability with thunk


e-schultz:
Currently both redux-observable and thunk are doing a `typeof action === 'function'` - was wondering if taking a cue from [redux-rxjs](https://github.com/acdlite/redux-rx/blob/master/src/observableMiddleware.js#L7) and checking if the action is an observable instead of just a function would allow for this to be compatible with thunk?

Or is there something else going on due to the support for teardown/cancelling an observable that I'm missing that wouldn't allow this to be possible?

jayphelps:
(not sure if you missed it or not, but this is [mentioned in the README](https://github.com/blesh/redux-observable#incompatible-w-redux-thunk))

There isn't any way to support both AFAIK because we can't check for Observable because that's not what you dispatch, you dispatch a function, just like redux-thunk dispatches functions, the difference being that the signature our middleware invokes that function with its `fn(actions: ActionsObservable, store: Store)` where as redux-thunk invokes the functions as `fn(dispatch: Function, getState: Function)`.  This factory style is crucial, because actions need to be able to receive an stream of all future actions, `ActionsObservable`, so that they can `takeUntil` (aka cancelling) and do other composition.

If you can think of an alternative that doesn't involve globals, do let us know.

jayphelps:
I guess it's technically possible we could switch to `fn(dispatch, getState, actions)` and then do `fn.length` to see if they provided 3 different args as our check, then swallow that action (just like today) instead of passing it along to redux-thunk. But since direct dispatch is totally an antipattern with redux-observable that's mostly out of the question just to support compatibility with redux-thunk. It's also rather brittle and ugly, obviously.

e-schultz:
True,

Was just thinking about existing applications that were using thunk, but I guess it wouldn't be too much work to refactor the action creators that were relying on dispatch to play nicely with this instead.

blesh:
We could get it to be a stand-in replacement for thunk if we changed the order and types of arguments we pass to the dispatched function. `(dispatch, getState, actions) => Observable...`

It would be a breaking change, of course. And we always return a subscription, so that wouldn't necessarily be compatible with existing thunk use

jayphelps:
Yeah, I mentioned that here:

[...]

PR #57, opened 2016-06-28
Thunk compatability concept

Igorbek:
This PR is just a concept that addresses #13
I wanted to see how it could work and sharing with you.

**Rationale**
`redux-thunk` is extremely popular middleware for `redux`, so that there're numerous working samples and boilerplates/starters that use it. Since `redux-observable` is incompatible with it, people need to choose one.

**Changes**
1. added ability to dispatch an observable itself, without wrapping it to a factory function
2. changed the way how a function being passed; instead of passing the function, a plain object that has `thunk` field containing the function, must be used
3. _bonus_: added `redux` module augmentation with dispatch call signatures

So again, I didn't propose that change in #13, so that is kinda proposal and a call to discussion.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,7 @@
-import { Middleware } from 'redux';
-import { Observable } from 'rxjs/Observable';
+import { Middleware, Action, Store } from 'redux';
+import { Observable, ObservableInput } from 'rxjs/Observable';
 import { Operator } from 'rxjs/Operator';
+import { AnonymousSubscription } from 'rxjs/Subscription';
 
 export declare function reduxObservable(): Middleware;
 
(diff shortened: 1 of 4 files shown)


jayphelps:
You might be happy to know we're very shortly deprecating the thunkservable-style in favor of the process manager style we introduced a little while ago but haven't yet documented. We're almost finished with the documentation around this new, preferred style in #55.

I'm thinking that instead of working around the thunkservables we'll just wait until we remove support for them and in parallel to introducing the deprecation warning also provide an config option to the middleware to remove support for them entirely so you can start using redux-thunk. Thoughts?

jayphelps:
You can view the in-progress docs here https://github.com/redux-observable/redux-observable/tree/6c3969e747878f979fb9637c117fd8394b2bb036/docs

Igorbek:
Oh, that's cool. I've just read the doc and it seems much more aligned with reactive concept. Are you going to remove dispatching observables completely in favor of process managers? Would it mean that even bare observables couldn't be dispatched?
I would suggest to introduce a separate middleware that would handle observables. It is still very useful, especially for it's cancellation capabilities.

jayphelps:
@Igorbek we plan to remove the support completely. You absolutely could introduce a new middleware to support dispatching observables directly! Please do link us to it here, if you do. 👍 

Regarding cancellation, we've found cancellation of the new process manager style to be more ideal, you dispatch a cancellation action that the process manager listens for with `.takeUntil(action$.ofType(CANCEL_TYPE))` or in some cases an automatic cancel with `.switchMap()` is ideal.

PR #70, opened 2016-07-14
fix(TypeScript type definition): Add combineEpics(), provide more accurate type info for others

### LINKED ISSUE #69: [TypeScript] combineEpics is missing in index.d.ts.
I got an TSC error using `redux-observable@0.7.0`

`(6,10): error TS2305: Module '"/path/to/project/node_modules/redux-observable/index"' has no exported member 'combineEpics`

Mostly because it's missing in index.d.ts

jayphelps:
fixes #69

@vire if you get a chance, could you confirm these new typings are good? I don't have any apps that use both TS and redux-observable yet. (though we do have a basic test of the typings included)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,25 +1,17 @@
-import { Middleware } from 'redux';
+import { Middleware, MiddlewareAPI, Action } from 'redux';
 import { Observable } from 'rxjs/Observable';
 import { Operator } from 'rxjs/Operator';
 
-export declare function createEpicMiddleware(): Middleware;
-
-// ./node_modules/rxjs/Observable.d.ts
-export declare class ActionsObservable<T> extends Observable<T> {
-
-  // ./node_modules/rxjs/Observable.d.ts specifies an `Observable`'s `source`
-  // property like so: `protected source: Observable<any>;`
-  // since `actionsSubject` is being assigned to an `Observable`'s `source`
-  // property it is being given the type: `Observable<any>`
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -1,25 +1,17 @@
...
-  // property it is being given the type: `Observable<any>`
-  constructor(actionsSubject: Observable<any>);
+export declare interface Epic {
+  (action$: ActionsObservable, store: MiddlewareAPI<any>): Observable<Action>;
+}
 
-  // ./node_modules/rxjs/Observable.d.ts specifies operators as generics like so
-  // `protected operator: Operator<any, T>;`
-  lift(operator: Operator<any, T>);
+export declare function createEpicMiddleware(...epics: Epic[]): Middleware;
```
vire: I'm a bit unsure about the spread epics here, I thought `createEpicMiddleware` takes just 1 epic (as result of combineEpics) if there are multiple like `createEpicMiddleware(epic1, epic2, epic3)` the epic2 and epic3 are just ignored?

vire:
I've updated my typings definiton from this PR and it works fine.

jayphelps:
@vire Thanks for confirming!

jayphelps:
@vire released as `0.7.1`

index.d.ts
jayphelps: Whoops, missed this comment. You're totally right. Going to correct it for a future release.

PR #72, opened 2016-07-15
docs(Epics): Add embedded JSBin examples in the Epics docs

jayphelps:
I looked for better embed solutions (that don't require a ton of setup) but couldn't find any. If someone has any recommendations on cleaner code embeds, please do let me know or even better PR 😍 

jsbin supports custom CSS styling (we could remove all the extraneous panels and fix line wrapping, etc) but it's only for paid customers and I am not.

### DIFF
--- a/docs/basics/Epics.md
+++ b/docs/basics/Epics.md
@@ -89,20 +89,23 @@ const pingEpic = action$ =>
     .mapTo({ type: 'PONG' });
 ```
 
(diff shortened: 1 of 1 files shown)


jayphelps:
FYI if you view the docs using HTTPS they won't be embedded directly, instead showing a "View this demo on JSbin" link. I'm looking into possibly signing up for a JSBin Pro account so we can serve them over HTTPS.

Issue #73, opened 2016-07-15
Hot Reloading?


ronag:
How does this work with hot reloading. Do we need some kind of `replaceEpic` similar to:

```javascript
  if (module.hot) {
    module.hot.accept('./reducers', () => {
      const reducers = require('./reducers').default
      store.replaceReducer(reducers)
    })
  }
```

jayphelps:
We're considering it #52 however Epics can maintain state, so if you hot reload them, that state disappears and I'm not confident it wouldn't put your app into a bad state because that state may actually rely being in sync or derived from the state in the store.

ronag:
How does cycles hot reloading get around that issue?

ronag:
Maintaining state in an Epic should be considered an anti-pattern, no?

ronag:
Though, I can see the problem with e.g. `debounceTime` where an action could disappear during a hot-reload.

jayphelps:
@ronag maintaining state is definitely _not_ a blanket anti-pattern. It depend on the state. Almost all side-effects require some sort of temporary state. Even a simple ajax call has state, inside the observable.

I can't provide any examples where Epic state being lost creates any real issues..but I imagine they exist at least in principle. Your `debounceTime` example certainly might be one of them, but I'm not sure if that's truly an issue in practice.

I don't know anything about Cycle's hot reloading, so I can't speak to it.

jayphelps:
A brief google search suggests to me hot reloading is provided by community modules like https://github.com/Widdershin/cycle-restart which apparently record and replay the actions. I'm not sure if something similar would be appropriate. If you replay actions into your Epics, that could kick off a whole mess of side effects like ajax calls, etc. That doesn't seem desirable and is potentially dangerous, if those ajax calls are not pure on the server.

ronag:
@jayphelps: I guess then that it is up to the developer to make sure that an epic is replayable using e.g. `Observable.cache`.

jayphelps:
Check out https://github.com/redux-observable/redux-observable/pull/75

ronag:
What if redux-observable sent a EOF action into the epics and the `replaceEpic` method used `concatMap` instead of `switchMap`?

jayphelps:
@ronag let's bring that sort of discussion over to #75.

jayphelps:
Released as `epicMiddleware.replaceEpic(nextEpic)`. Documented here: http://redux-observable.js.org/docs/recipes/HotModuleReplacement.html

[...]

PR #75, opened 2016-07-16
feat(replaceEpic): Added middleware method to replace the root Epic.

### LINKED ISSUE #52: Add ability to add new delegator

jayphelps:
**NOTE: I accidentally pushed an older version of [this to master days ago](https://github.com/redux-observable/redux-observable/commit/a8f458d1bf8db92c59a3e590869410e9ad73bf83) so #YOLO....part of this has already shipped. We can always revert, so it's not set in stone though.**

I'm still not convinced that using this is 100% safe, since Epics _may_ contain forms of state. There may be edge cases where losing that state causes the app to be in a bad state. For the most part that sort of state belongs in the store anyways, but this is likely not always obvious and there may be cases of ephemeral state that _should_ be in your Epic; after all, things like `Observable.ajax()` maintains internal state while the request is in-flight. Thinking we'll release it anyway and find out in practice whether it's an issue or not and then advise appropriately or rethink.

AFAICT redux-saga doesn't support Hot reloading either for basically the same reasons https://github.com/yelouafi/redux-saga/issues/22 Correct me if I'm wrong on this--the demos don't show it being used either, only for reducers.

Keep in mind that the old observables will all be unsubscribed, so you can use that opportunity to clean up any ephemeral state--things like `Observable.ajax()` will clean themselves up, cancelling in-flight AJAX requests. The `.finally()` operator is useful for this.

Closes #52, #53, #73

```js
import rootEpic from './where-ever-they-are';
const epicMiddleware = createEpicMiddleware(rootEpic);

if (module.hot) {
  module.hot.accept('./where-ever-they-are', () => {
    const rootEpic = require('./where-ever-the-are').default;
    epicMiddleware.replaceEpic(rootEpic);
  });
}
```

### DIFF
--- a/docs/basics/Epics.md
+++ b/docs/basics/Epics.md
@@ -1,7 +1,7 @@
 # Epics
 
 >##### Not familiar with Observables/RxJS v5?
-> redux-observable requires an understanding of Observables with RS v5. If you're new to Reactive Programming with S v5, head over to [http://reactivex.io/rxjs/](http://reactivex.io/rxjs/) to familiarize yourself first.
+> redux-observable requires an understanding of Observables with RxJS v5. If you're new to Reactive Programming with RxJS v5, head over to [http://reactivex.io/rxjs/](http://reactivex.io/rxjs/) to familiarize yourself first.
 
 An **Epic** is the core primitive of redux-observable.
 
--- a/package.json
+++ b/package.json
@@ -81,7 +81,6 @@
     "gitbook-cli": "^0.3.4",
     "json-server": "^0.8.14",
     "mocha": "^2.4.5",
-    "promise": "^7.1.1",
     "redux": "^3.5.1",
     "rimraf": "^2.5.2",
     "rxjs": "^5.0.0-beta.6",
--- a/src/EPIC_END.js
+++ b/src/EPIC_END.js
@@ -1,0 +1,1 @@
+export const EPIC_END = '@@redux-observable/EPIC_END';
--- a/src/createEpicMiddleware.js
+++ b/src/createEpicMiddleware.js
@@ -1,35 +1,44 @@
 import { Subject } from 'rxjs/Subject';
 import { BehaviorSubject } from 'rxjs/BehaviorSubject';
 import { from } from 'rxjs/observable/from';
+import { switchMap } from 'rxjs/operator/switchMap';
 import { ActionsObservable } from './ActionsObservable';
+import { EPIC_END } from './EPIC_END';
 
 export function createEpicMiddleware(epic) {
   const actionsSubject = new Subject();
   const action$ = new ActionsObservable(actionsSubject);
   const epic$ = new BehaviorSubject(epic);
+  let store;
 
-  const epicMiddleware = store => next => {
-    epic$.filter(epic => typeof epic === 'function')
-      .switchMap(epic => epic(action$, store))
-      .subscribe(store.dispatch);
+  const epicMiddleware = _store => {
+    store = _store;
 
-    return action => {
-      if (typeof action === 'function') {
-        if (typeof console !== 'undefined' && typeof console.warn !== 'undefined') {
-          console.warn('DEPRECATION: Using thunkservables with redux-observable is now deprecated in favor of the new "Epics" feature. See http://redux-observable.js.org/docs/FAQ.html#why-were-thunkservables-deprecated');
-        }
-
-        const out$ = from(action(action$, store));
-        return out$.subscribe(store.dispatch);
-      } else {
-        const result = next(action);
-        actionsSubject.next(action);
-        return result;
+    return next => {
+      if (typeof epic === 'function') {
+        epic$::switchMap(epic => epic(action$, store))
+          .subscribe(store.dispatch);
       }
+
+      return action => {
+        if (typeof action === 'function') {
+          if (typeof console !== 'undefined' && typeof console.warn !== 'undefined') {
+            console.warn('DEPRECATION: Using thunkservables with redux-observable is now deprecated in favor of the new "Epics" feature. See http://redux-observable.js.org/docs/FAQ.html#why-were-thunkservables-deprecated');
+          }
+
+          const out$ = from(action(action$, store));
+          return out$.subscribe(store.dispatch);
+        } else {
+          const result = next(action);
+          actionsSubject.next(action);
+          return result;
+        }
+      };
     };
   };
 
   epicMiddleware.replaceEpic = epic => {
+    store.dispatch({ type: EPIC_END });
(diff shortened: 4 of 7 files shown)


ronag:
What if `replaceEpic` send an `EOF`action and used `concatMap` instead of `switchMap`. That would allow the epic to be placed into a valid state before being replaced. Would solve e.g. debounce   cases.

jayphelps:
Side note, AFAICT redux-saga doesn't support Hot reloading either for basically the same reasons https://github.com/yelouafi/redux-saga/issues/22 Correct me if I'm wrong on this--the demos don't show it being used either, only for reducers.

jayphelps:
> What if replaceEpic send an EOF action and used concatMap instead of switchMap. That would allow the epic to be placed into a valid state before being replaced. Would solve e.g. debounce cases.

But then you're explicitly requiring all Epics `takeUntil` otherwise you'd duplicate Epics listening. That seems like a poor trade off to handle edge cases where replacing them might be a problem. Disagree?

ronag:
@jayphelps: Not necessarily. An `EOF` action might not even be required as long as epics are guaranteed to eventually end if `action$` completes.

So `replaceEpic` should `complete` the current `action$`, create a new stream from the new epic and `concatMap` map it on the old stream.

jayphelps:
@ronag I'm still _very_ hesitant. I can think of several cases in one of my app's codebase where the stream would not end because it's not derived from `action$`. I'd rather the default is everything you give me is completed, even if not derived from `action$`, as the new series are switched in. If someone wants to signal a shutdown, they certainly can before they `replaceEpic`.

```js
store.dispatch({ type: 'END' });
epicMiddleware.replaceEpic(nextEpic);
```

I feel this is the conservative approach, but I'm not writing off your suggestion.

Let's figure out some concrete cases where additional clean up is required. The `debounceTime` example I don't personally think is a big deal and arguably desired because the replacement epic may have different behavior.

ronag:
```javascript
(action$) => action$
  .ofType('UPDATE_SEARCH')
  .debounceTime(1000)
  .switchMap(x => API.search(x).map(actions.searchResult)
```

Would never get the search result, i.e. either FAILURE or SUCCESS and therefore the state `isSearching` would be invalid and the reducer is broken.

jayphelps:
@ronag 

Absolutely. I was thinking we'd advise in the documentation about this and recommend they dispatch some action as a signal--but perhaps we should indeed just dispatch one automatically? Is that what you were thinking?

```js
import { race } from 'rxjs/observable/race';

const epic = action$ =>
  action$.ofType('SEARCH')
    .debounceTime(1000)
    .switchMap(x => race(
      API.search(x)
        .then(result => ({ type: 'SEARCH_SUCCESS' }))
        .catch(result => ({ type: 'SEARCH_FAILED' })),
      action$.ofType('END')
        .mapTo({ type: 'SEARCH_CANCELLED' })
    ));


// later...
store.dispatch({ type: 'END' });
epicMiddleware.replaceEpic(nextEpic);
```
Anyone else have opinions or suggestions?

jayphelps:
FWIW redux-saga uses the `END` action as a signal to shutdown, used for SSR and (presumably) can be used for hot reloading.

> This release introduces a special action END which can be used to notify that an event source has terminated. You can also dispatch an END to notify Sagas that no more Store actions will be fired. This can be handy in Server Side rendering to break the while(true) loop inside watchers.The real-world example in the repo shows a possible ways to implement universal Sagas and SSR.

> END actions are automatically handled by the middleware and cause a Saga blocked on a take effect to automatically terminate (but it'll still wait for its forked tasks which provides support for SSR). To catch END values you can use the new Effect takem (aka takeMaybe). Using takem you get the explicit END so you can handle it manually. See #255 for more infos

ronag:
`@@EPIC_END` already exists in rxjs, `complete` you don't need any magical token.

jayphelps:
I'm not following. You example appears to still leave the store in a bad `isSearching: true` state. The reason I was suggesting an action to proceed completion is it's the only way you can emit another action to update the store to a good state.

ronag:
@jayphelps: Sorry, I fixed it now.

jayphelps:
@ronag I discussed this with several others who have been using redux-observable heavily at Netflix and they too believe completing the `ActionsObservable` and `concatMap` instead of `switchMap` isn't a safe solution. The biggest issue is that this _does not_ actually cancel anything that isn't derived directly from the `ActionsObservable`. That includes anything you `switchMap`/`flatMap` against an action. This is probably a bit counter-intuitive, but can be easily confirmed:

http://jsbin.com/gevijit/edit?js,output

```js
const subject = new Subject();
const action$ = new ActionsObservable(subject);

const intervalEpic = action$ =>
  action$.ofType('START_INTERVAL')
    .switchMap(() =>
      Observable.interval(1000)
        .mapTo({ type: 'INTERVAL' })
    );

intervalEpic(action$).subscribe(value => {
  console.log(value);
});

subject.next({ type: 'START_INTERVAL' });
// Observable.interval() does not derive from actions$
// so this will not stop it from emitting
subject.complete();
```
We've tested our the automatic emitting of a "end" action that you can listen for either in your Epics or obviously in your reducers as well. It seems to work well so far..but too early to know how good of solution it is in practice. We're thinking we'll merge that, let it settle in the community a bit, and then pivot if needed before v1.0. I've updated this PR with that code.

What are your thoughts on the issues I described to the `actions.complete()` solution and the proposed alternative?

Ultimately if someone wants to avoid this issue, they can avoid hot reloading (let normal full reloads act) and for async routes they can use pretty simple Rx to add new Epics without terminating the old ones:

http://jsbin.com/tuxikuk/edit?js,output

```js
const epic$ = new BehaviorSubject(someEpic);
const rootEpic = (action$, store) =>
  epic$.mergeMap(epic =>
    epic(action$, store)
  );

// sometime later...add another Epic, keeping the state of the old ones...
epic$.next(asyncEpic);
```
We might provide an API to do this like `runEpic` or `addEpic` but I'd prefer not to..I like keeping the middleware only deals with a single Epic and it's up to you to combine or merge how you wish.

None of this is "ideal" because dealing with side effects is not easy. as an aside, I'm truly surprised this isn't a bigger issue for redux-saga users historically. I'm guessing the answer is just that in practice it's not often a big deal because hot reloading/async route loading usually isn't done mid-UI interaction where state would be changing..also very few people load routes async.

jayphelps:
:shipit: If this ends up being bad, we're absolutely down to revisit. Want to move quickly and get people trying stuff.

PR #77, opened 2016-07-17
[DOC] Remove unused action$ Subject in test example

vire:
It's not used anywhere so I guess it's unnecessary...


jayphelps:
Good catch. Thank you!

PR #83, opened 2016-07-27
docs(WritingTests.md): Fix mockStore for each case

jhen0409:
Move `store = mockStore()` to `beforeEach(...)`.


jayphelps:
good catch! 👍

PR #85, opened 2016-07-30
feat(Adapters): Adds support for Epic input/output adapters.

jayphelps:
This allows us to support RxJS v4 via https://github.com/redux-observable/redux-observable-adapter-rxjs-v4

An adapter is an object that has two properties, `input` and `output`.

```js
const adapter = {
  input: action$ => convertItSomeHow(action$),
  output: action$ => convertItSomeHow(action$)
};
```

The input function is provided the original v5 `ActionsObservable` and the result you return is then provided to all of your Epics.

The output function is invoked for each Epic, provided with the output of that Epic. The result you return is expected to be RxJS v5 compliant; anything that `Observable.from` works on so that includes `stream[Symbol.observable]()`.

We strongly believe people should switch to v5, but we realize that not everyone has the immediate cycles to do so, or have issues against software labeled as "beta"--though it's funny they would still use redux-observable (or similar middleware) which _is far more unstable than v5 is_.

Cc/ @blesh @BerkeleyTrue

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@
 ## Install
 
 This has a peer dependencies of `rxjs@5.0.*` and `redux`, which will have to be installed
-as well. **It is not currently compatible with RxJS v4**
+as well. Support for RxJS v4 can temporarily be provided by [an optional adapter](https://github.com/redux-observable/redux-observable-adapter-rxjs-v4).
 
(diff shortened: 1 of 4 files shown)


jayphelps:
Published as v0.9.0

Issue #87, opened 2016-07-31
Add more recipes


jayphelps:
Everyone,

Let's add a bunch of thoughts here about Recipes people would like to see added to the docs. If you can, do in fact feel free to just go ahead and PR any you think of or that are on this list! 🚀 

- [ ] Router navigation ([related](https://github.com/reactjs/react-router-redux#what-if-i-want-to-issue-navigation-events-via-redux-actions))
- [ ] Logging analytics (like gaq, etc)
- [ ] Patterns with common libs like redux-form ([related](https://github.com/yelouafi/redux-saga/issues/161#issuecomment-191312502))
- [ ] Best practices/anti-patterns

gusvargas:
Do you think these examples should be view library/framework agnostic? For example, should the router navigation examples actually use `react(-redux)?(-router)?` or do you imagine that we'd just simulate routing related actions similar to what `react-redux-router#push` may generate while showcasing interesting epics.

jayphelps:
Both 😄

I think it's important to try and keep things unbiased and present the general patterns that should apply to all frameworks. But because there will be subtleties in the real world so we should also try and provide specific examples for at least the three or four frameworks too (like react-router requiring some integration with redux)

winkerVSbecks:
@jayphelps It might be helpful to add a polling recipe. Is this the best approach or can you recommend something better? 
https://jsbin.com/bupeja/52/edit?js,output

jayphelps:
@winkerVSbecks polling is a great recipe! I would recommend placing the ajax stuff inside the first switchMap so that it has access to the original `action.payload`:

https://jsbin.com/xiwasu/edit?js,output

I'm thinking the recipe for that should prolly be a different use case than fetching a user. Maybe something like logging or checking internet connection.

Timopheym:
I'll be glad to see one about websocket streams support... ;)

murraybauer:
@jayphelps fetching a 'newsfeed' could be a good example

cloud-walker:
For the polling example I suggest using `exhaustMap` instead of `mergeMap` to ensure promise resolve

jayphelps:
@cloud-walker care to elaborate? As an aside, I don't recommend using promises for these sorts of situations unless you have no control over the originating API--because Promises aren't cancellable and aren't lazy.

cloud-walker:
I've made an [example](http://jsbin.com/rohopon/edit?js,console) without promises, to clarify my point.

[...]

PR #89, opened 2016-08-02
chore(examples): Add navigation examples

gusvargas:
Related to https://github.com/redux-observable/redux-observable/issues/87

First stab at some navigation examples. It's a small app that let's you search for GH users. Clicking on a GH user gives you links to their repos.

The epic examples are:

1. Side loading data by responding to navigation actions
2. Canceling the side loading of data due to an action originating from a navigation
3. Restricting routes/delayed redirects

I'm relatively new to Rx so I'd appreciate any input on how to write more idiomatic use of operators. Also happy to work in any other ideas for navigation-related epics.


jayphelps:
This is rad. Will dig into it when I have a moment 👍

examples/navigation/containers/Admin.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+export default function Admin() {
+  return (
+    <p>
+      You don't have access. Redirecting back home.
+    </p>
+  );
+}
```
blesh: You'll want to rename these files to `.jsx`, if for no other reason than to get GitHub support for JSX formatting.

examples/navigation/epics/adminAccess.js
```diff
@@ -0,0 +1,15 @@
...
+function hasAccess() {
+  return false;
+}
+
+export default function adminAccess(action$) {
+  return action$.ofType(LOCATION_CHANGE)
+    .map(action => action.payload.pathname)
+    .filter(pathname => pathname === '/admin')
+    .map(hasAccess)
+    .filter(access => !access)
```
blesh: I don't understand the purpose of these two lines:

```js
.map(hasAccess) // hasAccess always returns false.
.filter(access => !access) // ...so this will allow everything through
```

They seem like they can be omitted.

examples/navigation/epics/clearSearchResults.js
```diff
@@ -0,0 +1,11 @@
...
+import * as ActionTypes from '../actionTypes';
+
+export default action$ =>
+  action$.ofType(LOCATION_CHANGE)
+    .filter(({ payload }) =>
+      payload.pathname === '/' && !!!payload.query.q
+    )
+    .mapTo({
+      type: ActionTypes.CLEARED_RESULTS
+    });
```
blesh: Something to note here about `mapTo`, it will literally emit the same instance of the object over and over. If anything downstream is checking to see if the action is a new instance, it could have consequences. In contrast, `map(() => { foo: 'bar' })` will emit a new object instance each time.

examples/navigation/epics/fetchReposByUser.js
```diff
@@ -0,0 +1,36 @@
...
+    fetchIntents$
+      .switchMap(user =>
+        ajax.getJSON(`https://api.github.com/users/${user}/repos`)
+          .map(res => ({
+            type: ActionTypes.RECEIVED_USER_REPOS,
+            payload: {
+              repos: res,
+              user
+            }
+          }))
```
blesh: It's slightly more performant to use the mapping function that is the second argument to `switchMap`:

```js
switchMap((user) => ajax.getJSON(`https://api.github.com/users/${user}/repos`),
                   (user, res) => ({
                     type: ActionTypes.RECEIVED_USER_REPOS,
                     payload: {
                       repos: res,
                       user
                     }
                   }))
```

examples/navigation/index.js
```diff
@@ -0,0 +1,30 @@
+import 'rxjs';
```
blesh: This is interesting. Why are we just script-importing `rxjs` here?

examples/navigation/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "main": "index.js",
+  "scripts": {
+    "serve": "webpack-dev-server --inline --history-api-fallback",
+    "build": "webpack"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/redux-observable/redux-observable.git"
+  },
+  "author": "",
```
blesh: BRO. Your name here.

examples/navigation/epics/adminAccess.js
gusvargas: Sorry, I should have left a comment. Those dummy lines are to imply that more meaningful code should go there to *actually* check for access – it's just for semantics. I could leave a comment to the existing code or just omit it all together.

examples/navigation/epics/clearSearchResults.js
gusvargas: Ah – I never considered that that anyone would care about the identity of actions. Good point.

examples/navigation/epics/fetchReposByUser.js
jayphelps: @blesh I don't believe we want to advocate that for these sort of circumstances because it will confuse them and create bugs when they want to add error handling. e.g. if they put a `.catch()` that returns an error action inside the switchMap, now will wrongfully be mapped as the success response. If they place that `.catch()` outside the switchMap, it is too late and will prevent their Epic from listening to future actions.

examples/navigation/index.js
jayphelps: @blesh it is done so that all the operators are placed on `Observable.prototype` because Epics receive an existing observable that they do not create. This is suggested in the docs: http://redux-observable.js.org/docs/basics/Epics.html#a-basic-example

gusvargas: This is originally why I added it – the operators I wanted were not on the `ActionsObservable` and I was immediately sure which operators I was going to want to use. However, I just removed it and now *Just Works*.

jayphelps: @gusvargas it shouldn't work.... o.O

jayphelps: @gusvargas OH it works because you're effectively doing the same thing in `searchUsers.js`

https://github.com/redux-observable/redux-observable/pull/89/files/82ef94d41941f5d35b6bdd7a834bd13ffdc7165f#diff-98f4ddf4496646ae41dc030ac31bb6b5R1

You're importing the root `rxjs` to get `Observable`..the root `rx` imports _every_ operator and puts them on the prototype.

I would keep the `import 'rxjs';` in the index.js because it's the safest way to say "the app gets all of RxJS"

examples/navigation/epics/adminAccess.js
blesh: I would omit the lines of code, but add a comment telling the reader what they _could_ do there.

examples/navigation/index.js
gusvargas: Ah – I see. Thanks for clearing that up.

examples/navigation/epics/fetchReposByUser.js
blesh: >  if they put a .catch() that returns an error action inside the switchMap, now will wrongfully be mapped as the success response

I can only assume you mean they're trying to handle some error that could have occurred during the mapping of the response? If so, then you're right.

I'm mostly indifferent, but *I personally* would like people to know about the projection function on `switchMap`, `flatMap`, `concatMap`, etc.

Then again, when/if I ever get the sync operations flattening to land, it shouldn't matter as much. There will still be an added closure, which will slow things down, but it will work.

I retract my suggestion. Mostly because Jay cares more than I do. haha.

examples/navigation/index.js
blesh: I see. That makes sense. I'd add a comment though, because it reads funny to those of us who are used to only importing exactly what we need from rxjs.

gusvargas:
@jayphelps @blesh I've addressed all the comments so far. Any other suggestions?

examples/navigation/index.js
gusvargas: Comment added in https://github.com/redux-observable/redux-observable/pull/89/commits/e885d0ad45f73b9c9405a9044a029a179a3f11cb

blesh:
I think this example is sizing up to be great. I feel like I've learned things.

blesh:
LGTM :shipit:

gusvargas:
Had some side conversation with @jayphelps about a couple things, mainly decoupling data fetching from the router action and having the intent come form the container component. I'll take make those changes soon.

blesh:
> Had some side conversation with @jayphelps about a couple things

Oh... it would be good for you two to try to document/summarize these sorts of discussions on the issues next time (no need this time).

jayphelps:
@blesh agreed. I usually do this, just didn't because the discussion was lengthy and forgot to give a short summary.

tl;dr for numerous lengthy reasons we agreed that the epics should listen for a component dispatched actions instead of `LOCATION_CHANGE`.

blesh:
> tl;dr for numerous lengthy reasons we agreed that the epics should listen for a component dispatched actions instead of LOCATION_CHANGE.

That's a bummer, because that was the interesting thing I learned from this example. :\

jayphelps:
@blesh maybe you'll convince me otherwise 😄

gusvargas:
These are the highlights of that discussion in case anyone else has something to add:

- Coupling the URL to the epic adds yet another place that would need to be updated if paths were ever to change.
- In the case of `react-router-redux` it doesn't hook into `react-router` so you have to roll your own code for extracting path params.
- You can't be consistent with the pattern because not every container component's data needs are parameterized in the URL.
- However, it is really nice that there is this implicit machinery that understands when data should be fetched.

One alternative would be for container components to explicitly request their data needs by firing off an action. This approach comes with a boilerplate issue, i.e., both `componentDidMount` and `componentWillReceiveProps` would need to be implemented to capture all cases where data may need to be fetched. This is boilerplate that could be abstracted away but that's outside the scope of this PR.

Neither of us are certain that one pattern is better than the other but we're leaning towards the latter since it seems like even without an abstraction to reduce the boilerplate it will still result in less brittle/inconsistent code than the former.

gusvargas:
@jayphelps @blesh I updated this PR per the points listed above.

jayphelps:
Started reviewing, will finish soon. So far it looks super sexy. 😎

jayphelps:
🎉 Huge thank you @gusvargas !!

blesh:
Awesomeness, @gusvargas! 🍻

Issue #96, opened 2016-08-16
implicit return-type error during typescript compilation


KyleCartmell:
When I try to compile my typescript project which uses redux-observable, the following error messages result:

node_modules/redux-observable/index.d.ts(7,3): error TS7010: 'lift', which lacks return-type annotation, implicitly has an 'any' return type.
node_modules/redux-observable/index.d.ts(8,3): error TS7010: 'ofType', which lacks return-type annotation, implicitly has an 'any' return type.

KyleCartmell:
https://github.com/redux-observable/redux-observable/pull/97

jayphelps:
I read #97 and I'm not sure what the status of these two are.

Is there still a problem? could you summarize what the issue is? I see several things mentioned and that PR closed so it's not clear what the issue truly is.

KyleCartmell:
Issue above persists. When I compile a TypeScript project which uses redux-observable tsc complains about implicit 'any' return types. Adding return type information to the typings file exposed by redux-observable resolves the error. The PR I submitted did the needful but used the wrong types. I see your comment in the PR comments stating that ActionsObservable is the correct type, even for ofType. I'll happily open a new PR with the correct types if you feel that this issue is clear and that the types are desirable.

PR #97, opened 2016-08-16
added explicit return type definitions

KyleCartmell:
added explicit return type definitions to life and ofType member functions within ActionsObservable class (github issue #96)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -4,8 +4,8 @@ import { Operator } from 'rxjs/Operator';
 
 export declare class ActionsObservable extends Observable<Action> {
   constructor(actionsSubject: Observable<Action>);
-  lift(operator: Operator<any, Action>);
-  ofType(...key: any[]);
+  lift(operator: Operator<any, Action>) : ActionsObservable;
+  ofType(...key: any[]) : ActionsObservable;
 }
 
 export declare interface Epic {


KyleCartmell:
I am unable to reproduce this build failure locally (my local build works perfectly with and without the changes in this PR) and the error output from the log isn't great. I'd be interested to see if the issue is repeatable but I don't see a way to re-run the build. Can someone take a look at this for me?

KyleCartmell:
Ahh, there is a problem here. The return type on ofType definitely shouldn't be boolean, it's an observable of some sort. I'm puzzling through precisely what types would be best exposed here.

Not sure that this problem is the cause of the code 2 return from tsc in the CI build; I get some awesome error messages out of the test harness now that I'm invoking it correctly.

I'll close this and try to do something smarter, but feedback would still be greatly appreciated.

index.d.ts
```diff
@@ -4,8 +4,8 @@ import { Operator } from 'rxjs/Operator';
 
 export declare class ActionsObservable extends Observable<Action> {
   constructor(actionsSubject: Observable<Action>);
-  lift(operator: Operator<any, Action>);
-  ofType(...key: any[]);
+  lift(operator: Operator<any, Action>) : ActionsObservable;
+  ofType(...key: any[]) : boolean;
```
gusvargas: https://travis-ci.org/redux-observable/redux-observable/builds/152735720#L388

```
test/typings.ts(10,6): error TS2339: Property 'mapTo' does not exist on type 'boolean'.
```

`ofType` also returns `ActionsObservable`, not `boolean`. The boolean returns you see in [`ofType`](https://github.com/redux-observable/redux-observable/blob/master/src/ActionsObservable.js#L16-L30) are for returning from the `filter` predicate.

gusvargas: Oops. You beat me to finding the problem. 👍

KyleCartmell: I think that lift should explicitly return ActionsObservable but I'm not sure about ofType.
(full disclosure: I'm still trying to grok RxJS.)

ofType does a filter thing, which returns an Operator, but Operator is Observable, right? So... explicitly returning Operator doesn't seem wrong but I'm not sure that's semantically correct.

jayphelps: Both `lift` and `ofType` return `ActionsObservable`, though indeed that's not currently annotated.

jayphelps:
@KyleCartmell reopened based on our discussion in #96. Please amend your existing commit with the fix--it also needs to follow our [commit message formatting requirements](https://github.com/redux-observable/redux-observable/blob/master/CONTRIBUTING.md)

KyleCartmell:
@jayphelps corrected for review

Thanks for the help and lesson!

jayphelps:
LGTM. Thanks @KyleCartmell !

jayphelps:
@KyleCartmell noticed you cut your own build so wanted to let you know that 0.9.1 was released and includes this fix. 🍻

KyleCartmell:
Sadly, I'm stuck on 0.8.0 until https://github.com/angular/angular/issues/9549.

For anyone else trying to follow this thread later, I discovered after I'd pulled the declaration into the 0.8.0 version that I could have (and now have) just added the the declaration to the typings mechanism in angular2-seed which feeds into tsc and does the needful. (I'm also a version behind on angular2-seed (2.0.0-rc4) so good luck, future humans, if you're trying to do this thing!)

PR #99, opened 2016-08-30
Update Troubleshooting.md

kokokenada:


jayphelps:
@kokokenada

> follow the docs and:

Can you add clarification to it about when this happens? We don't document any usage of Epics as classes, so I assume this is probably related to third party documentation about Angular2 usage?

i.e. the example doesn't explain why you would be using a class to begin with so it would be confusing for users who just use Epics as functions.

kokokenada:
Sure. I changed the first sentence to read "If you are organizing your epics into a class. (E.g. in order to benefit from Angular 2 dependency injection), you might have made the mistake of using class methods:"

jayphelps:
woot! thank you @kokokenada

Issue #56, opened 2016-06-27
RFC: provide state$ as a stream argument


jayphelps:
The store provided by redux (and given to Epics) isn't a full store, so it doesn't have the `store[Symbol.observable]()` interop point to support `Observable.from(store)`. [This is by their design](https://github.com/reactjs/redux/issues/1833).

Let's consider changing the Epic signature to `function (action$: ActionsObservable, state$: BehaviorSubject<State>, store: Store)`, making a stream of stage changes the second argument and moving the store to the third argument.

``` js
const fetchUserEpic = (action$, state$) =>
  action$.ofType(FETCH_USER)
    .mergeMap(action =>
      getJson(`/users/${action.id}`, { 'Authorization': `Bearer ${state$.value.authToken}` })
        .map(respose => fetchUserFulfilled(response))
    );

// or the "reactive" way, but more verbose

const fetchUserEpic = (action$, state$) =>
  action$.ofType(FETCH_USER)
    .withLatestFrom(state$.pluck('authToken'))
    .mergeMap(([action, authToken]) =>
      getJson(`/users/${action.id}`, { 'Authorization': `Bearer ${authToken}` })
        .map(respose => fetchUserFulfilled(response))
    );
```

``` js
const localStorageStateEpic = (action$, state$) =>
  state$
    .filter(state => state.someState)
    .distinctUntilChanged() // state is supposed to be immutable, so this should be safe
    .throttleTime(500) // may or may not want to do something like this for perf reasons
    .do(state => localStorage.setItem('someState', JSON.stringify(state)))
    .ignoreElements(); // we have no actions to emit
```
I'd want to keep the store still, since while not necessarily idiomatic to use, is a handy escape hatch to "get shit done".

If this is a good idea in general, we'd need to decide what kind of observable the state changes were in. My initial thoughts were a `BehaviorSubject` so that the imperative `value` property was available for the same common cases people use `store.getState()` for or they can operate on it and it emit the last value right away. ~~Because it's an observable, it feels slightly more idiomatic and then people don't need to learn about the fact that Redux's store supports the `Symbol.observable` interop so you could just do `Observable.from(store)`, which isn't immediately obvious.~~

Obviously, this is not really all that different than existing solution..it's more whether we should provide a obvious "Rx way" or not. Please feel free to debate this.

Cc/ @blesh

blesh:
The only thing that wouldn't be great about this is the fact that you'd remove the ability to `dispatch` from a `do()` block.

```js
[...]

PR #100, opened 2016-09-02
Using Store enhancer instead of middleware - PoC

tomkis:
As it was mentioned in #56 using middlewares has its limitations. I believe, we should ideally switch to enhancer ASAP as this is obviously a breaking change (yet not that painful).

Please, this is just a proof of concept, we'd need to adjust tests etc. I've just filed the PR so that we can discuss whether it would even make sense.


jayphelps:
@tomkis1 can you provide a list of pros vs cons to switching to a store enhancer?

mcoetzee:
@jayphelps From my recent experiments, a pro is that you can pass the actual store to the epics, or extend the API to include Observable.from(store) or something like that.

rgbkrk:
The epics get the store as the second argument.

```
export const myEpic = (action$, store) => 
  ...
```

jayphelps:
@rgbkrk the store provided is actually not the _real_ store. It doesn't have `store[Symbol.observable]()` so you cannot consume it as an Observable with `Observable.from(store)` like you can the real store.

This is an intentional limitation by redux, one they [do not wish to change](https://github.com/reactjs/redux/issues/1833).

rgbkrk:
_Gasp_. Thank you.

tomkis1:
My apologies, totally forgot about this PR.

There's just one pro which is obviously improved flexibility in implementation since with store enhancer, you can do hypothetically anything (even replace the entire redux implementation). It will also open the doors for any future improvements. Solves https://github.com/redux-observable/redux-observable/issues/56


The obvious drawback is that it's not that easy to use, since as a user, if you want to use multiple enhancers (even `applyMiddleware` is enhancer!) you have to somehow `compose` it and you need to know in what order (this is partially issue even for middlewares), therefore you need to understand how it works.

Sawtaytoes:
Is this DoA or is it still a possibility?

jayphelps:
@Sawtaytoes it's not off the table, but I've been piloting an alternative solution to state$-as-a-stream #56 that doesn't require switching to an enhancer. People are far less familiar with enhancers so I think it would be ideal if we didn't switch to one.

Sawtaytoes:
I agree, I think that's a far better solution. I had created a subspace-like solution because having `store` as the second argument requires mocking `store` in unit tests rather than a simple `getState()` function. After seeing that other solution, working with `state$` gets around having store the contents of the `getState()` function and can lead back to the functional implicit returning epics.

jayphelps:
@Sawtaytoes if you give state$ a shot, lmk in the coming weeks what you think of it. 👍

jayphelps:
Closing due to inactivity. If this needs to be re-opened, let me know. Thanks!

PR #105, opened 2016-09-09
fix(typings): Add generics so developer defines what redux Actions look like

jayphelps:
Based on discovery and conversation of this issue on Gitter: https://gitter.im/redux-observable/redux-observable

You will now can pass in the expected shape of redux actions

```ts
interface FluxStandardAction {
  type: string | symbol | any;
  payload?: any;
  error?: boolean | any;
  meta?: any
}

const rootEpic = combineEpics<FluxStandardAction>(epic1, epic2);
const epicMiddleware = createEpicMiddleware<FluxStandardAction>(rootEpic);
// later

// types/generics inside epics can be inferred automatically!
const someEpic = (action$, store) =>
  action$.ofType(SOMETHING)
    .map(({ payload }) => ({
      type: ELSE,
      payload
    }));
```

Or of course you can use `any` by just not providing them

```ts
const epicMiddleware = createEpicMiddleware(rootEpic);
```

Cc/ @SethDavenport

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ lib
 temp
 dist
 _book
+typings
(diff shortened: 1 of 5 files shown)


jayphelps:
Published as 0.10.0

PR #106, opened 2016-09-11
refactor(thunkservables): Removed support for thunkservables

### LINKED ISSUE #13: Compatability with thunk

jayphelps:
BREAKING CHANGE: Support for thunkservables to be removed, replaced by Epics. After this, you may now use redux-thunk in tandem with redux-observable.

I need to update the FAQ page before merging this, going to make a new entry for http://redux-observable.js.org/docs/FAQ.html#why-were-thunkservables-removed

Closes #13 as well.

Cc/ @blesh

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -26,6 +26,9 @@ Not out of box, but we provide an adapter [redux-observable-adapter-rxjs-v4](htt
 It should be possible for you to create an adapter to add support. See [redux-observable-adapter-rxjs-v4](https://github.com/redux-observable/redux-observable-adapter-rxjs-v4) for reference.
 
 <a id="miscellaneous-thunkservables-deprecated"></a>
-### Why were thunkservables deprecated? 
+<a id="why-were-thunkservables-removed"></a>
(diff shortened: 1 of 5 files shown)


BerkeleyTrue:
@jayphelps How long before this version is released?

jayphelps:
@BerkeleyTrue day or two. You have actually been able to use redux-thunk by placing it _before_ the epic middleware in the `applyMiddleware` order. That way, redux-thunk will consume the function before it has a chance to reach redux-observable.

jayphelps:
@BerkeleyTrue went ahead and cut a release now. `0.11.0`

blesh:
:shipit:

PR #109, opened 2016-09-12
Fix broken docs links in README

joshburgess:
The links to the docs pages currently do not work if trying to access them via the main repository @ https://github.com/redux-observable/redux-observable


jayphelps:
I really don't want to hardcode the full urls to these. Makes it hard to refactor/rename/etc and makes local running of the docs harder.

I'll look into how we can resolve the conflict between github vs. website. (you're welcome to as well, of course)

istarkov:
Looks like it's github bug, relative links from the lists does not work. One option is to remove lists syntax.

BerkeleyTrue:
@joshburgess Thanks. It looks like those links have since been removed.

Happy Coding

PR #111, opened 2016-09-14
fix(test): feed multiple actions through

rgbkrk:
I'm guessing this is what the test was intended for, to send multiple actions through and make sure they come out on the other side.

### DIFF
--- a/test/ActionsObservable-spec.js
+++ b/test/ActionsObservable-spec.js
@@ -12,11 +12,11 @@ describe('ActionsObservable', () => {
 
   it('should support ActionsObservable.of(...actions)', () => {
     const output = [];
-    const action$ = ActionsObservable.of({ type: 'FIRST', type: 'SECOND' });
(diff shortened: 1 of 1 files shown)


jayphelps:
😱 Indeed, that was the intent. Good catch! I slightly modified your commit message because we don't want test changes (even fixes) to show up in the CHANGELOG we autogenerate.

Thanks @rgbkrk !!

rgbkrk:
Fine by me, the format change makes sense - otherwise this would shoe up as fixing something in the core lib. Still getting used to this format when I send patches to here and RxJS. Does this also mean long commit messages are cool here?

jayphelps:
@rgbkrk I don't think it requires any official length, but as a best practice we typically keep the first line fairly short as a summary, then if needed add a more in-depth description below it, paragraph style. We're not super strict about about the past-tense grammar part of things, but it's [explained here](https://github.com/bcoe/conventional-changelog-standard/blob/master/convention.md#commit-message-format), in case you hadn't come across this.

PR #115, opened 2016-09-20
feat(combineEpics): combineEpics() now transparently passes along all arguments

jayphelps:
which means now you can now easily inject dependencies to all your epics:

```js
// given epics like this
const epic1 = (action$, store, api) => {
  // access to API here!
};
```

```js
import * as api from './api/location';

// inject API into all epics as third arg
const rootEpic = (...args) => combineEpics(epic1, epic2)(...args, api);
```

which makes it easier to test:

```js
const fakeApi = {
  fetchSomething: () => Observable.of('fake response')
};
epic1(action$, fakeStore, fakeApi);
```

We may provide this functionality as an option to the middleware at some point since this composition may not be immediately obvious to people.

```js
// HYPOTHETICAL ONLY
createEpicMiddleware(rootEpic, { dependencies: [api] });
```

### DIFF
--- a/src/combineEpics.js
+++ b/src/combineEpics.js
@@ -3,5 +3,7 @@ import { merge } from 'rxjs/observable/merge';
 /**
   Merges all epics into a single one.
  */
-export const combineEpics = (...epics) => (actions, store) =>
-  merge(...(epics.map(epic => epic(actions, store))));
+export const combineEpics = (...epics) => (...args) =>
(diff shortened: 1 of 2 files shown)


BerkeleyTrue:
OMG yes.

I literally updated my local version just this second to work on this.

jayphelps:
v0.12.0

PR #120, opened 2016-09-24
Fix Error

seijiro:
I faced below errors in example programs.

I think it don't need  "..." operator.

please check this.

```
$ npm run serve

ERROR in ./reducers/reposByUser.js
Module build failed: SyntaxError: Unexpected token (7:8)

   5 |     case ActionTypes.REQUESTED_USER_REPOS:
   6 |       return {
>  7 |         ...state,
     |         ^
   8 |         [action.payload.user]: undefined
   9 |       };
  10 |     case ActionTypes.RECEIVED_USER_REPOS:

```

### DIFF
--- a/examples/navigation/reducers/reposByUser.js
+++ b/examples/navigation/reducers/reposByUser.js
@@ -3,15 +3,13 @@ import * as ActionTypes from '../ActionTypes';
 export default function reposByUser(state = {}, action) {
   switch (action.type) {
     case ActionTypes.REQUESTED_USER_REPOS:
-      return {
-        ...state,
+      return Object.assign({}, state, {
         [action.payload.user]: undefined
-      };
+      });
     case ActionTypes.RECEIVED_USER_REPOS:
-      return {
-        ...state,
+      return Object.assign({}, state, {
         [action.payload.user]: action.payload.repos
-      };
+      });
     default:
       return state;
   }


jayphelps:
Actually, making that change makes it no longer do as indended. It's using object spread syntax.


https://babeljs.io/docs/plugins/transform-object-rest-spread/

```js
  switch (action.type) {
      case ActionTypes.REQUESTED_USER_REPOS:
        return {
 -        ...state,
 +        state,
          [action.payload.user]: undefined
        };
      case ActionTypes.RECEIVED_USER_REPOS:
        return {
 -        ...state,
 +        state,
          [action.payload.user]: action.payload.repos
        };
      default:

**Comment by jayphelps** [comment]
Actually, making that change makes it no longer do as indended. It's using [object spread syntax](https://babeljs.io/docs/plugins/transform-object-rest-spread/) which requires using a transpiler like Babel w/ that transform enabled.

We probably shouldn't be using this experimental syntax in our example repo though because not everyone will be familiar with it and have support for it. You can use `Object.assign()` to achieve the same thing:

```js
switch (action.type) {
  case ActionTypes.REQUESTED_USER_REPOS:
    return Object.assign({}, state, {
      [action.payload.user]: undefined
    });
  case ActionTypes.RECEIVED_USER_REPOS:
    return Object.assign({}, state, {
      [action.payload.user]: action.payload.repos
    });
  default:
    return state;
}
```

I can update our examples to use `Object.assign`, or you're welcome to as well.

jayphelps:
I didn't notice you mentioned this doesn't work even inside the example itself using `npm run serve` from it.

It works fine for me, can you confirm you've run `npm install` first? Either way I think it's a good idea to use `Object.assign` instead for examples.

seijiro:
I tried ``` npm install ``` first. However faced same error.

Thank you for  good idea.
I'll replace ``` ... ``` to ``` Object.assign ``` syntax on local code.

ex.
it is executed command list

```
$ cd workspace/redux-observable
$ cd example/navigation
$ npm install
$ npm run build
$ npm run serve
```

```
npm version 3.10.3
node v6.4.0
```

jayphelps:
I updated your [commit message](https://github.com/redux-observable/redux-observable/pull/120/commits/e0ada96ceab737926231ef0bc8cb40c4ee753fd4) to match our commit [convention requirements](https://github.com/redux-observable/redux-observable/blob/master/CONTRIBUTING.md#sending-a-pull-request). Take a peek for future PRs.

Thanks!

PR #122, opened 2016-09-26
chore(examples): restructure epics for better clarity

joshburgess:
- [ ] I have read the [Contributor Guide](https://github.com/redux-observable/redux-observable/blob/master/CONTRIBUTING.md)
- [ ] My commit messages are in [conventional-changelog-standard](https://github.com/redux-observable/redux-observable/blob/master/CONTRIBUTING.md#sending-a-pull-request) format.


gusvargas:
Hey @joshburgess would you mind providing some context into why these changes provide better clarity? Within some margin clarity can be a pretty subjective thing.

I'll just give my 2-**bikesheddy**-cents for why I think breaking out may not be super valuable.

1. It seems to a add another layer of indirection without good reason at (what seems like) an arbitrary boundary. What I mean by this is, conceptually it's not hard to buy that a `searchUsers` saga will encapsulate the implementation details of responding to a `SEARCHED_USER` action, i.e., as a reader I can give `searchUsers` a shallow glance and grok its role. I think there *can* be benefits to breaking the saga up into a piece that just debounces the signal and a piece that just performs the side effect, e.g., perhaps there are other sagas that depend on on the debounced signal. In this case however, there is no other saga that depends on the output of the `searchUsersDebounced` saga, that's what I mean by addition another layer of indirection without a good reason.

Obviously, there is benefit in modularity and breaking larger computations into smaller ones – especially for the purpose of clarity. In this case, however, placing that split boundary at `debounceTime` seems arbitrary. The `searchUsers` flow as many parts, i.e., filtering by action type, filtering out empty queries, debouncing, performing the side effect, and last mapping to some success action. So why break up the flow at the debouncing stage vs any of the other stages? Again, I think there would be a benefit here if there were multiple downstream sagas that consumed on the output of `searchUsersDebounced` but that doesn't seem to be the case.

2. Semantically it doesn't seem great to leak side-effect implementation details to constructs outside of the side-effect facilitator(s). This could easily be resolved by making the `searchUsers` (and `SEARCHED_USERS` action type) the "public" API while letting `SEARCHED_USERS_DEBOUNCED` be more of a "private" saga-to-saga action. This would also be more aligned with the locally (in this example) established past tense action type convention, i.e., the action type described what has happened, not what will happen. That convention is, of course, [up for debate](https://github.com/reactjs/redux/issues/384) but it's what's in place currently in this example.

Please let me know if there are any considerations that I over looked. Sorry for all the bike shedding but "clarity" related changes are prone to summoning it.

examples/navigation/epics/adminAccess.js
```diff
@@ -8,9 +8,9 @@ export default function adminAccess(action$) {
...
     .delay(2000)
-    .mergeMap(() => Observable.merge(
-      Observable.of(accessDenied()),
-      Observable.timer(2000)
-        .map(() => push('/'))
-    ));
+    .mergeMap(() =>
+      Observable.of(push('/'))
+       .delay(2000)
+       .startWith(accessDenied())
```
gusvargas: The reason I was partial (when originally writing up the example PR) to the `merge` option is because it reads well top to bottom. I'm no Rx ninja so for all I know using the `startWith` is more idiomatic. If that's the case I wonder what's more clear to beginners (considering that many devs will be introduced to Rx via `redux-observable`).

I don't really have a lot of experience teaching devs rx but maybe @jayphelps or @blesh can shed some light here.

PR #127, opened 2016-10-04
fix(UMD build): now properly sets up external requires statements

### LINKED ISSUE #126: UMD export not working properly with System.js
Specifically, if you look at the [third and sixth lines in the current umd file](https://unpkg.com/redux-observable@0.12.0/dist/redux-observable.js)

You'll notice it's requiring `undefined`.

```js
module.exports = factory(require(undefined), require(undefined), require(undefined));
```

```js
exports["ReduxObservable"] = factory(require(undefined), require(undefined), require(undefined));
```

blesh:
After speaking with @TheLarkInn, it seemed that what was coming from `webpack-rxjs-externals` was incomplete. This PR removes that dependency and adds the stuff we're using specifically.

fixes #126

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,6 @@
     "sinon": "1.17.6",
     "typescript": "^2.0.3",
     "typings": "1.4.0",
-    "webpack": "^1.13.1",
-    "webpack-rxjs-externals": "~0.0.3"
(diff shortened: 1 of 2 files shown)


blesh:
Closing in favor of updating `webpack-rxjs-externals`.

PR #128, opened 2016-10-04
fix(UMD): bump webpack-rxjs-externals to correct UMD generation

jayphelps:
fixes #127

Upstream webpack-rxjs-externals change: https://github.com/jayphelps/webpack-rxjs-externals/pull/1

Cc/ @blesh 

(it was already matching under our semver range, but bumping to reduce likelyhood of local caching of incorrect older lib and have a clear trace back to the fix)


jayphelps:
Released as v0.12.1

PR #133, opened 2016-10-14
refactor(rxjs): no longer use function-bind syntax

jayphelps:
This is an internal detail, users should not be affected.

No longer use function-bind syntax, use private Symbols for Observable operators we use internally. We don't want to continue depending on a stage-0 syntax. The reason we did was we don't want to decorate the `Observable.prototype` with operators because then users of redux-observable can then accidentally depend on the fact that a operator we need internally exists on that prototype. So if we no longer use it, that fact could cause users apps to break because they aren't importing the operator as well.

With the private Symbols approach, we add the operators we need for internal use but since they're Symbols that are not public, it's not possible for users to accidentally depend on them.

Neither of these approaches will be necessary in the future if RxJS gets something like: https://github.com/ReactiveX/rxjs/pull/2034

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,7 +1,6 @@
 {
     "presets": ["es2015"],
     "plugins": [
-      "transform-function-bind",
       "transform-es2015-modules-commonjs",
       "transform-object-rest-spread"
     ]
--- a/package.json
+++ b/package.json
@@ -69,7 +69,6 @@
     "babel-eslint": "^7.0.0",
     "babel-loader": "^6.2.4",
     "babel-plugin-transform-es2015-modules-commonjs": "^6.11.5",
-    "babel-plugin-transform-function-bind": "^6.8.0",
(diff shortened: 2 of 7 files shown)


BerkeleyTrue:
Why not just use the `.call` syntax. That avoids the need for Symbols and also reduces the amount of code being added.

jayphelps:
@BerkeleyTrue certainly not the end of the world today

```js
switchMap.call(
  map.call(epic$, epic => epic(action$, store)),
  action$ => adapter.output(action$)
).subscribe(store.dispatch);
```

I hate it, but it's not _that bad_. The main beef I have with it is that I have a bunch of experimental additions (server rendering, forking short lived epics, etc) that also need some operators and after 2 operators this gets pretty old. Perhaps it's OK.

evertbouw:
The bind operator was removed in https://github.com/redux-observable/redux-observable/pull/403 so I think we can close this

Issue #58, opened 2016-06-28
Feature Suggestion: Accept an object in combineManagers


matthewwithanm:
Like `combineReducers`, something like this 

```js
if (managers.length === 1 && typeof managers[0] !== 'function') {
  const managersObj = managers[0];
  managers = Object.keys(managersObj)
    .filter(key => typeof managersObj[key] === 'function')
    .map(key => managersObj[key]);
}
```

This would allow a pattern like:

```js
// ActionManagers.js
export const incrementIfOddManager = ...;

// elsewhere
import * as ActionManagers from './ActionManagers';
const managers = combineManagers(ActionManagers);
```

jayphelps:
Is this to support the shorthand `import * as ActionManagers from './ActionManagers';` but no other benefit, right? I'm not seeing any harm in adding this, so I'm initially 👍 . Any caveats you can think of?

matthewwithanm:
Yup, just for that. I can't think of any caveats other than it won't really help with larger apps where your managers are spread throughout multiple files.

BLamy:
If combined epics is passed an object personally I would like to see the keys used as type filters.

Instead of having something like this:
```js
const attemptConnectEpic = action$ =>
  action$.ofType(ATTEMPT_CONNECT)
     ::switchMap(attemptConnectPromise)
     ::map(didConnectAction);

const didConnectEpic = action$ =>
  action$.ofType(DID_CONNECT)
     ::switchMap(getUserInfoPromise)
     ::map(didGetUserInfoAction);

combineEpics(attemptConnectEpic, didConnectEpic);
```

I'd like to see something like this
```js
const attemptConnectEpic = attemptConnect$ =>        
  attemptConnect$::switchMap(attemptConnectPromise)::map(didConnectAction);

const didConnectEpic = didConnect$ =>
  didConnect$::switchMap(getUserInfoPromise)::map(didGetUserInfoAction);

combineEpics({
    [ATTEMPT_CONNECT]: attemptConnectEpic, 
    [DID_CONNECT]: didConnectEpic
});

```

jayphelps:
@BLamy while that's handy for the most common cases, I'm initially hesitant since I've noticed more and more people doing

```js
import * as epics from './epics';
const rootEpic = combineEpics(epics);
```

[...]

PR #134, opened 2016-10-24
docs(SettingUpTheMiddleware): Clarify arguments

DannyDelott:
Just set up my rootEpic and stumbled a little bit here. There is a subtle but important difference between the `combineEpics` and `combineReducers` signatures.

### DIFF
--- a/docs/basics/SettingUpTheMiddleware.md
+++ b/docs/basics/SettingUpTheMiddleware.md
@@ -6,7 +6,7 @@ Now that we know what [Epics](Epics.md) are, we need to provide them to the redu
 
 Just like redux requiring a single root Reducer, redux-observable also requires you to have a single root Epic. As we [learned previously](Epics.md), we can use `combineEpics()` to accomplish this.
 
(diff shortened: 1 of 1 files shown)


jayphelps:
I think #58 is looking more attractive. i.e. actually supporting an object, even though it doesn't make a lot of sense if you think about it (the keys are meaningless, unlike with reducers). I keep seeing this issue pop up because people are so familiar with `combineReducers` and doing things like


```js
import * as epics from 'whatever';

const rootEpic = combineEpics(epics);
```

DannyDelott:
That's an interesting discussion.

I'm inclined to agree with you about the object parameter being awkward given the lack of an implicit hierarchy in the epics.

Also, since there is no implicit order to the epics passed as arguments to `combineEpics`, then you don't need an object, since things like `params` and `options` arguments are often used solely to obsolete argument order.

But I see the point about doing what folks are used to. This was just a little bit of a hunt during my first setup of redux-observable and a callout would have been nice. All around awesome library!

jayphelps:
Going to go ahead with #58 Thanks again for your PR! I really do appreciate it.

PR #137, opened 2016-10-27
chore(errors): detect epics which do not return anything

jayphelps:
Provides a more descriptive error when someone accidentally doesn't return something from their epics

```js
const someEpic = () => {
  action$.ofType(SOMETHING).etc(); // whoops! forgot to return the stream!
};
```

Someone may ask why we don't try and check whether what they returned is an Observable e.g. `output instanceof Observable`. The reason is that we support any stream-like thing that RxJS supports, which is Observable, an array, an iterable/generator, or object which has `Symbol.observable` property. [see here](https://github.com/ReactiveX/rxjs/blob/master/src/util/subscribeToResult.ts). I think checking for this exhaustive list is excessive, because RxJS itself will check and error too (assuming it's not undefined, which is what we're protecting against now)

I'll try and update RxJS to also provide a more descriptive error when passed `undefined`, but since this happens seemingly often enough we'll still keep this check in redux-observable too. Right now RxJS assumes it's at least an object, so it errors out when it reaches the iterable check `Cannot read property 'Symbol(Symbol.iterator)' of undefined`

### DIFF
--- a/src/combineEpics.js
+++ b/src/combineEpics.js
@@ -5,5 +5,11 @@ import { merge } from 'rxjs/observable/merge';
  */
 export const combineEpics = (...epics) => (...args) =>
   merge(
-    ...epics.map(epic => epic(...args))
+    ...epics.map(epic => {
+      const output$ = epic(...args);
+      if (!output$) {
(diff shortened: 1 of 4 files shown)


rgbkrk:
We had a contributor that added on to an epic that used the fat arrow bind with an expression and once it was a statement (in forgetting a return), it was not obvious where the issue was. This will be super helpful for folks!

PR #150, opened 2016-11-17
feat(typings): Adds type definitions for ActionsObservable.from/of

jayphelps:


index.d.ts
```diff
@@ -1,8 +1,27 @@
...
+  static of<T>(item1: T, scheduler?: Scheduler): ActionsObservable<T>;
+  static of<T>(item1: T, item2: T, scheduler?: Scheduler): ActionsObservable<T>;
+  static of<T>(item1: T, item2: T, item3: T, scheduler?: Scheduler): ActionsObservable<T>;
+  static of<T>(item1: T, item2: T, item3: T, item4: T, scheduler?: Scheduler): ActionsObservable<T>;
+  static of<T>(item1: T, item2: T, item3: T, item4: T, item5: T, scheduler?: Scheduler): ActionsObservable<T>;
+  static of<T>(item1: T, item2: T, item3: T, item4: T, item5: T, item6: T, scheduler?: Scheduler): ActionsObservable<T>;
+  static of<T>(...array: Array<T | Scheduler>): ActionsObservable<T>;
+
+  static from<T>(ish: ObservableInput<T>, scheduler?: Scheduler): ActionsObservable<T>;
+  static from<T, R>(ish: ArrayLike<T>, scheduler?: Scheduler): ActionsObservable<R>;
```
jayphelps: The existing method overrides we have are required because Rx doesn't yet use lift (or something else) to derive from our `ActionsObservable`. These type signature overrides are required because otherwise our method overrides would return the generic `Observable` type instead of `ActionsObservable`, since that's what the inherited definitions return.

PR #151, opened 2016-11-26
feat(ActionsObservable): Support for .ofType(predicate)

Andarist:

As heavy `redux-saga` user and collaborator I've found this pattern really useful in making some generic sagas, not focused on exact `type` for example. String comparison is sometimes just not enough. 

Im happy to fix PR if you have some stylistic suggestions (or any other too!). 

If you accept the change, I'm gonna write tests and docs and include them in the PR via amend. 

Also I guess most of the use cases (but not all of them) can be easily accomplished with `filter` in userland, but even for those its nice (I think) to have this behaviour builtin.

### DIFF
--- a/src/ActionsObservable.js
+++ b/src/ActionsObservable.js
@@ -24,9 +24,13 @@ export class ActionsObservable extends Observable {
   }
 
   ofType(...keys) {
-    return this::filter(({ type }) => {
+    return this::filter((action) => {
+      const { type } = action;
       const len = keys.length;
       if (len === 1) {
+        if (typeof keys[0] === 'function') {
+          return keys[0](action);
+        }
         return type === keys[0];
       } else {
         for (let i = 0; i < len; i++) {


rgbkrk:
Thanks for the PR!

I'm a bit confused about how this isn't equivalent to using `filter` directly:

```js
import 'rxjs/add/operator/filter';

action$.filter(predicate)
```

Especially since that's what `keys[0](action)` is doing, evaluating the predicate with the action.

Andarist:
Okay, I've kinda misunderstood process of creating epics. You are right ofc. I somehow got into my head that this is possible - `action$.ofType(ACTION_TYPE).filter(predicate)`, didnt quite grasp from quick scan of the lib that `action$` is a stream itself so direct `filter` on it is valid as well.

rgbkrk:
🤔 Perhaps it's worth it for `redux-observable` to have an example that uses another Rx operator so people know they can compose the stream however they want. No matter what, this library takes for granted that people are determined to get past the Rx learning cliff (at least some of it) -- I'm still wrapping my head around ways to compose these. Having a blast so far!

One other thing I do with inner streams/observables is use `takeUntil` with another type for cancellation: `...takeUntil(action$.ofType('CANCEL_REQUEST'))`.

Issue #64, opened 2016-07-11
reduxObservable.js Uncaught TypeError: Cannot read property 'subscribe' of undefined


casoetan:
Hi.

This maybe me, but I keep getting this error

<code>
reduxObservable.js:21 Uncaught TypeError: Cannot read property 'subscribe' of undefined
</code>

My manager

```js
import Rx from 'rxjs'
import { todoHz } from '../db' // I'm using horizon.io

import {
  getTodosSuccessful
} from '../actions/todos'

const { ajax, of } = Rx.Observable;

const todosManager = action$ => {
  action$.filter(action => console.log('action:', action))
  console.log(action$)
  action$
    .ofType('request todos')
    .merge(todoHz.fetch())
    .map(todos => {
      console.log('todos:', todos)
      getTodosSuccessful(todos)
    })
    .catch(err => console.log(err))
}

export default todosManager
```

this is then combined (as I intend to have different managers in different files going forward)
```js
import { combineDelegators } from 'redux-observable'

import todosManager from './todos'

export default combineDelegators(todosManager)
```

My store

```js
import { createStore, applyMiddleware, compose } from 'redux'

import invariant from 'redux-immutable-state-invariant'
import devTools from 'remote-redux-devtools'

import { reduxObservable } from 'redux-observable';
import { persistStore, autoRehydrate } from 'redux-persist'

import localforage from 'localforage'

import { actionLogger } from '../middleware'

import rootReducer from '../reducers'
import rootManager from '../managers'

export default function configure(initialState) {
  const middleware = [invariant(), reduxObservable(rootManager), actionLogger]

  const enhancer = compose(
    applyMiddleware(...middleware),
    autoRehydrate(),
    devTools({
      name: 'todo_devtool',
      realtime: true,
      hostname: 'localhost',
      port: 8000,
      maxAge: 30,
      filters: { blacklist: ['EFFECT_RESOLVED'] }
    })
  )

  const store = createStore(rootReducer, initialState, enhancer)

  localforage.config({
    name: 'todo_dev',
    version: 1.0,
    storeName: 'todo_dev_app', // Should be alphanumeric, with underscores.
    description: 'Todo Dev'
  })

  persistStore(store, {
    storage: localforage
  })

  if (module.hot) {
    module.hot.accept('../reducers', () => {
      const nextReducer = require('../reducers')
      store.replaceReducer(nextReducer)
    })
  }

  return store
}
```

Can you help me.

[...]

PR #160, opened 2016-12-12
Validate epic returns Subscribable

kwonoj:

related to #64.

This PR updates `combineEpic` to have naïve check to given epic is `Subscribable`, by checking if it has `symbol` or either simply has `subscribe` to allow subscribe to it. Latter is for use cases of user using adapter and using non-Rx based which doesn't have symbol yet.

In addition to those, simplified test scripts does not require separate build for test specs by allowing mocha directly compiles it.


jayphelps:
Sorry for the delay, was on vacation. I don't believe we need to test for symbol.observable because they would only be using this `combineEpics` if they are using RxJS v5. If they are using anything else with an adapter, they would need to use their own equivalent. Which reminds me that the adapters I maintain should probably include their own `combineEpics` since the default one won't work with anything but rxjs v5.

kwonoj:
I see. feel freely close PR if isn't feasible.

jayphelps:
@kwonoj if we do want to enforce it be an observable, I don't mind the subscribe check! the existing is an even more naive check because I'm somewhat reluctant to not support the same signature as `Observable.from`. I don't really have a use case for peoples epics returning an array/promise/etc but I could see someone being overly clever with a generator...not that I would recommend that. I dunno. it's almost certainly ok. I just need to stew on it a bit.

kwonoj:
I think in general someone experienced with `redux-observable` probably fine without this PR, as they certainly know epic should return observable. It's just helping initial dive in, including myself there was really simple errors to forget return observable and spend several hours to figure out why code is not working. So, it may helpful for someone but not obviously benefitable for everyone, reason I'm ok if you don't see feasibility feel free to close.

jayphelps:
@kwonoj I'm very interested in this. What were you returning that passed a truthy check but wasn't an observable?

jayphelps:
@kwonoj I _really_ want these sorts of check for newcomers, not for the people already familiar. So your insight will help guide.

kwonoj:
@jayphelps apologize I don't even recall what was I experienced exactly since I experienced last year(!) and did not leave any memory for it. :/

Anujan:
@jayphelps  For newcomers, a common error may be something like this?

`catch(error => ({ action: 'ACTION_FAILED' }))`

jayphelps:
Closing for now, I think if we were going to do a more robust check it would need to be all the things that Rx supports, Observable/Promise/Iterable/etc

kwonoj:
Yup, got it.

PR #161, opened 2016-12-15
Update Cancellation.md

incrediblesound:
docs(typo): removed superfluous "after" from "after inside our .mergeMap()"


rgbkrk:
Thanks!

PR #165, opened 2016-12-22
docs(navigation-example): request via nextProps

rgbkrk:
@gusvargas - was this meant to do a query based on the `user` from the `nextProps` if the old props didn't match?

### DIFF
--- a/examples/navigation/containers/ReposByUser.jsx
+++ b/examples/navigation/containers/ReposByUser.jsx
@@ -8,13 +8,6 @@ class ReposByUser extends React.Component {
     this.props.requestReposByUser(this.props.params.user);
   }
 
-  componentWillReceiveProps(nextProps) {
-    const { user } = this.props.params;
-    if (user !== nextProps.params.user) {
-      this.props.requestReposByUser(user);
-    }
-  }
-
   render() {
     const {
       reposByUser,


gusvargas:
Yep. Interesting.. You ran this locally and didn't see the correct behavior? I'm just wondering how this could've checked out when I wrote it.

rgbkrk:
I wasn't running it locally, I was using it as a reference while working on a project and wondered about this.

gusvargas:
Ah. I just ran this locally to refresh my memory. The reason why I never observed any wrong behavior is because there is no way for the `user` to change while `ReposByUser` is rendered, i.e., `ReposByUser` doesn't expose any interactions that would cause `user` to update. So really, the whole `componentWillReceiveProps` hook can be removed.

Sorry about that.

rgbkrk:
> Sorry about that.

Hey that's ok. This is our only example so I'm leaning heavily on it. Should I remove this function instead then?

While I'm super comfortable with Rx and redux-observable, I'm moderately uncomfortable about it seems like I have to fetch data on mount with React router. In the case of the app I'm working on, I need ~4 API calls to be made - three that are for general data for the overall app and one for the specifically requested id (that is passed via `props.params.id`). Really hoping I can grok the right patterns here.

gusvargas:
Yeah I think it'd be a good idea to just totally remove that lifecycle method in this PR since it's unnecessary. 

Here is some more discussion around the decision to kick off data fetching from the lifecycle method: https://github.com/redux-observable/redux-observable/pull/89#issuecomment-239307469

FWIW, I work on a fairly large redux + redux-observable + react-router + react-router-redux app and originally I did all my data fetching implicitly within epics that respond to `react-router-redux` navigation actions. I started moving the firing of my data fetching actions into lifecycle methods because it's annoying having to roll your own/duplicate url param parsing logic while also keeping it in sync with what your routes define. All of that parsed data is already immediately available to you within route handler components.

Currently my root `Application` component (in `componentDidMount`) just fires one action per "global" data that is needed app wide and blocks further rendering until that data has been resolved. Eventually I'd like to create a HoC that abstracts that boilerplate away, i.e., firing the actions, knowing what to do on success, failure, etc.

rgbkrk:
Ok great I'll update the PR later today.

Your commentary on your app is super helpful, I'll be coming back to refer to it later today.

jayphelps:
@rgbkrk is this ready?

rgbkrk:
This is good to go, thanks for the reminder.

Issue #71, opened 2016-07-14
Pass function to ofType


nicholaswyoung:
When a function (like the ones returned by [redux-actions](/acdlite/redux-actions) `createAction()`) are passed to `ofType()`, they should be converted via `toString()` automatically. String action specifiers would be passed onward, unmodified.

Current syntax:

``` javascript
const queueSomethingEpic = createAction('app/activities/epic')
const doSomethingEpic = action$ => action$.ofType(queueSomethingEpic.toString())
```

Proposed syntax:

``` javascript
const doSomethingEpic = action$ => action$.ofType(queueSomethingEpic)
```

This seems to be an emerging method in the react/redux community. I've found it a useful pattern. But does it belong in redux-observable?

jayphelps:
@nicholaswyoung can you provide some additional examples of where this pattern is becoming popular? in redux-actions, the `actionFactory.toString()` behavior appears undocumented AFAICT

jayphelps:
Also--if we supported this either with explicit `.toString()` or casting `'' + type` wouldn't we be breaking anyone who wants to use things other than strings, for types? Support for that was added to redux, like using Symbols.

Can you think of a way of maintaining both, without any side effects?

jayphelps:
btw--you can inject this behavior pretty easily. You could even publish it as an npm package if you want to. A sort of redux-observable "plugin".

```js
// patch-redux-observable.js
import { ActionsObservable } from 'redux-observable';

const originalOfType = ActionsObservable.prototype.ofType;
ActionsObservable.prototype.ofType = function ofType(...types) {
  return originalOfType.apply(this, types.map(type => type.toString()));
};
```

Then import it in, as side effects in some common entry file, such as where you `createStore()`.

```js
import { createEpicMiddleware } from 'redux-observable';
import './patch-redux-observable';

// etc such as createStore()
```

nicholaswyoung:
@jayphelps I'm fairly sure the OP in #10 is using a similar technique. It also looks like a [hot topic](https://github.com/reactjs/redux/issues/628) over on the redux issue tracker as well. I picked up the idea from @jeffbski, who also turned me on to this library. 

All that to say, I seem to be uniquely positioned among folks who favor this technique, perhaps to the point of existing inside my own echochamber.

The more I think about it, this probably does belong as a plugin. At least for the time being, while we're feeling out the benefits and pitfalls of this technique.

jayphelps:
[...]

PR #175, opened 2017-01-19
feat(ActionsObservable.ofType): allow passing action creators with a toString

danielfarrell:
Use case for this is using the [redux-actions](https://github.com/acdlite/redux-actions) repo to make action creators and using those instead of constants in the ofType.

Biggest win is in the case of following a ducks pattern and not wanting to export the constants.  You can see a super simplified action creator in the test on this branch as an example.

### DIFF
--- a/src/ActionsObservable.js
+++ b/src/ActionsObservable.js
@@ -27,10 +27,10 @@ export class ActionsObservable extends Observable {
     return this::filter(({ type }) => {
       const len = keys.length;
       if (len === 1) {
-        return type === keys[0];
+        return type === keys[0].toString();
       } else {
         for (let i = 0; i < len; i++) {
-          if (keys[i] === type) {
+          if (keys[i].toString() === type) {
             return true;
           }
         }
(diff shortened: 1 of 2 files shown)


jayphelps:
This was discussed in #71. I'm still concerned about this breaking support for using Symbols as types.

```js
var type = Symbol();
type.toString() === type;
// false!
```

We could also call toString on the action type as well, but then we have the problem of unrelated Symbols would now match against each other

```js
var a = Symbol();
var b = Symbol()
a.toString() === b.toString()
// true (not good)
```

danielfarrell:
Gotcha, sorry I missed that.  I did think about it and looked to make sure Redux does not support symbols as action types(because it breaks replay/serialization: https://github.com/reactjs/redux/issues/4).

jayphelps:
@danielfarrell that thread is outdated, redux now supports it. https://github.com/reactjs/redux/pull/295

danielfarrell:
Makes sense.  And yeah, symbols should be supported, you shouldn't merge this.  Closing.

jayphelps:
Thanks for being cool! 🎉 

At the expense of stating the obvious, you could just call toString while passing it to `ofType` 😄

PR #191, opened 2017-02-10
chore(rxjs): update rxjs to v5 non-beta

jayphelps:
BREAKING CHANGE:

RxJS v5 non-beta (e.g. 5.1.0) is now required. Upgrading from rxjs 5
beta to latest should be easy in most cases.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 
 ## Install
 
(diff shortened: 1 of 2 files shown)


rgbkrk:
Hurrah!

Issue #163, opened 2016-12-20
[Feature request] Inject dependencies in epics


AugustinLF:
In my app when I don't use redux-observable, I use a variant of redux-thunk that lets me inject dependencies in my actions (see how it is used in this [action](https://github.com/este/este/blob/master/src/common/todos/actions.js#L5)), it's also something made possible by the [redux-inject](https://www.npmjs.com/package/redux-inject) middleware. This lets me have efficient DI in my actions, depencencies set up at the store creation. However, redux-observable doesn't provide such an API.

For instance, it would work like that
```javascript
const store = createStore(
  rootReducer,
  applyMiddleware(epicMiddleware, dependencies),
);
```

Would that be something you would be interested? If so, I'd gladly work on a PR.

jayphelps:
Are you wanting to inject things into the **actions** themselves or inject things into your **epics**? 

I feel pretty confident we won't support injecting things into the actions themselves--it's really outside the scope of redux-observable and easily done with other middleware you would in addition.

For injecting dependencies into the epics, this is already supported as a pattern, but no first-class API exists for it yet. See #115

``` js
// given epics like this
const epic1 = (action$, store, api) => {
  // access to API here!
};
```

``` js
import * as api from './api/location';

// inject API into all epics as third arg
const rootEpic = (...args) => combineEpics(epic1, epic2)(...args, api);
```

which makes it easier to test:

``` js
const fakeApi = {
  fetchSomething: () => Observable.of('fake response')
};
epic1(action$, fakeStore, fakeApi);
```

AugustinLF:
Thanks for your answer, I indeed meant injecting things into the epics! I didn't think of your workaround, that would do the job, great :)

But your proposal is what I had in mind.

bali182:
If this is implemented, how would additional arguments be handled? Would the config objects `dependencies` field simply overtake the role of passing more stuff into epics or would the current way of passing arguments kept, and dependencies passed after what is currently passed?

jayphelps:
@bali182 if you mean would this replace

```js
const rootEpic = (...args) => combineEpics(epic1, epic2)(...args, api);
```

yes, it would. You could still do this, but it would be bumped to the fourth arg position instead of the third, as the third would be anything injected using the `dependencies` option. If you don't inject anything using `dependencies`, I'm not sure if we would choose to either force `null` or `undefined`, or just not pass anything at all, freeing up that parameter position.

[...]

PR #193, opened 2017-02-11
Added explicit dependencies option to createEpicMiddleware

bali182:
Implements #163 


bali182:
@jayphelps let me know if I should add more tests or change something!

jayphelps:
Wow awesome!! Impressed with the test coverage actually including the discussed combineEpics spread support.

Everything looks great, we just gotta decide if defaulting to `null` is the right choice instead of not passing anything at all. Any arguments either way?

bali182:
Of course, I wasn't sure about `null` either, it could be `undefined` as well. I didn't really want to do anything opinionated like an empty object or array, but I'm open to change it to any other default value.

BerkeleyTrue:
@bali182 Nice work. I'd definitely think leaving it undefined is a better option.

@jayphelps should the function avoid passing in the third option if it undefined?

jayphelps:
@BerkeleyTrue I believe so, that way `arguments.length` is not 3 (which it is even if you pass `undefined` literally.) Kinda a pain, but I think it's the "right thing to do".

I've amended the PR with those changes. Can someone give a final review before I merge?

jayphelps:
Actually, I think we should be a stickler about not merging this until it includes documentation, even if brief. I think another section under Basics, something like "Injecting Dependencies Into Epics" but @bali182 might have better suggestions?

bali182:
@jayphelps Sounds good as a title, roughly that's how I googled it and found the open issue :)

If you want I could help with the docs, have some free time this afternoon.

jayphelps:
@bali182 please do! can just amend this PR

bali182:
@jayphelps Done! Please do look at [createEpicMiddleware.md](https://github.com/redux-observable/redux-observable/pull/193/files#diff-a7715f80ded1874f5f0084ebddf948d3) with extreme care, because I added some description about the `options` argument, and I'm not 100% sure you planned it this way. Let me know I should change it anywhere!

jayphelps:
@bali182 awesome. I made some changes, can you give me your honest thoughts? https://github.com/redux-observable/redux-observable/pull/193/commits/e83a27393be76264eaa08c0fd60abc59e641fe9d

docs/recipes/InjectingDependenciesIntoEpics.md
```diff
@@ -1,55 +1,79 @@
...
+
+const fetchUserEpic = (action$, store) =>
+  action$.ofType('FETCH_USER')
+    .mergeMap(({ payload }) =>
+      ajax.getJSON(`/api/users/${payload}`)
+        .map(response => ({
+          type: 'FETCH_USER_FULFILLED',
+          payload: response
+        }))
+    );
```
jayphelps: I mostly changed this to fetching a single user (simpler as an example to grok) and adjusted some things to be consistent with the other doc examples like mergeMap vs flatMap, isolating the observable chains by putting the map inside the mergeMap.

jayphelps: I mostly changed this to fetching a single user (simpler as an example to grok) and adjusted some things to be consistent with the other doc examples like mergeMap vs flatMap, isolating the observable chains by putting the map inside the mergeMap.

bali182:
@jayphelps Read the whole thing, thanks for correcting it (especially the `Observable => any` and `any => Observable`, this is exactly why I asked to take a look :)! The recipe looks much clearer, having an actual example of how things can be mocked helps a lot. Also it helps that the wording is coming from a native English speaker. I'd say it's good. Only thing I don't understand is that why `mergeMap` is used in all examples. Coming from java/scala `flatMap` is much more self explanatory for me but that's just personal preference.

jayphelps:
@bali182 thanks for reading.

Some backstory: mergeMap is preferred because that's what RxJS v5 prefers. The documentation, links, filenames, etc are all mergeMap, only referencing that flatMap is an alias. The justification of that change in RxJS is obviously not redux-observable related, but was primary because we discussed the naming of flatMap + flatMapLatest with the community and by and large people (who were not already familiar with the term flatMap) preferred mergeMap and that it was quicker and easier to understand what it did. _That said_...that decision was made like almost 2 years ago and since then casually discussed the possibility of going back to flatMap but sticking with switchMap (instead of flatMapLatest). But I wouldn't bet on it changing.. For now we just go with the canonical name RxJS v5 uses.

bali182:
@jayphelps thanks for the explanation! Didn't know about the `mergeMap` preference in rxjs

BerkeleyTrue:
Noice!

jayphelps:
Released in 0.14.0 https://github.com/redux-observable/redux-observable/blob/master/CHANGELOG.md#0140-2017-03-02

bali182:
Great! :tada: :tada: :tada:

Issue #84, opened 2016-07-28
redux-epic and redux-observable


BerkeleyTrue:
Hello Everyone,

I'm the creator of Redux-Epic, which is a redux library in similar fashion to redux-saga but built using RxJs with first class support for SSR.

In https://github.com/BerkeleyTrue/redux-epic/issues/13 it has been brought to my attention that this library has arrived at the same path as redux-epic, triggering observable in middleware a la redux-saga.

But there are also various differences that I go into detail in the issue above.

I'm opening this issue to discuss whether merging the two products would benefit everybody and whether the goals I've laid out for redux-epic are compatible with the goals of redux-observable.
- First class support for SSR
- Designed for Universal JavaScript (read: Isomorphic)
- Ability to inject dependencies
- Disposable Epics

jayphelps:
Nice! I'll provide some of my initial thoughts.

Because we both use the term Epic, but seem to use them differently, for the purpose of the following Epic will **always** mean redux-observable's concept of Epic, which I believe is still called Saga in redux-epic?

### First Class Support for Server Side Rendering
We'd love this, but have complex concerns. I can't immediately tell if they are solved by redux-epic.

Summary: We will have to spend more time around this. Our concerns may end up unfounded.

### Universal JavaScript First Design
I think this is the same thing as First Class Support for Server Side Rendering?

### Dependency injection
Looking at the docs, I'm not sure how this is implemented in redux-epic so I can't comment on it. I imagine it's similar to being able to inject a third param to all Epics. Often an API library or something, so they can easily mock it in testing.

```js
const fetchUserEpic = (action$, store, api) =>
  action$.ofType(FETCH_USER).etc();
```
We've been experimenting with exactly this internally. If it continues to work well, we do plan to add it or obviously the community could PR this ability before hand.

This is a pretty obvious solution, so I imagine it's very similar to the one you landed one too.

Summary: We are experimenting with patterns that may eventually become part of the library to try and minimize API churn.

### RxJS v4 instead of RxJS v5
### Once RxJS is out of beta and has fully replaced Rx then this library will move over.
I believe this one is the Number 1 practical difference between the two, in my opinion.

[...]

PR #197, opened 2017-02-16
fix(typings): Add combineEpics overload for custom Epic signatures

inuscript:

I try to use API injection epic pattern that like https://github.com/redux-observable/redux-observable/issues/84#issuecomment-236042600 and #163.

But current typing is not accept `combineEpics` with extra argument epics.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -35,7 +35,7 @@ export interface EpicMiddleware<T, S> extends Middleware {
   replaceEpic(nextEpic: Epic<T, S>): void;
 }
 
-interface Adapter { 
+interface Adapter {
   input: (input$: Observable<any>) => any;
   output: (output$: any) => Observable<any>;
 }
@@ -47,3 +47,4 @@ interface Options {
 export declare function createEpicMiddleware<T, S>(rootEpic: Epic<T, S>, options?: Options): EpicMiddleware<T, S>;
 
 export declare function combineEpics<T, S>(...epics: Epic<T, S>[]): Epic<T, S>;
+export declare function combineEpics<E>(...epics: E[]): E;
--- a/test/typings.ts
+++ b/test/typings.ts
@@ -1,5 +1,5 @@
 import { expect } from 'chai';
-import { createStore, applyMiddleware } from 'redux';
+import { createStore, applyMiddleware, MiddlewareAPI } from 'redux';
 import { Observable } from 'rxjs/Observable';
 import { asap } from 'rxjs/scheduler/asap';
 import 'rxjs/add/observable/of';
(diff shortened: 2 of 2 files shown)


jayphelps:
hmmm this would lost type safety on those extra args since they're all `any`.

What about instead creating an overload for `combineEpics` which allows you pass in **your own** implementation of the Epic interface, so you can have safety.

```ts
declare function combineEpics<T, S>(...epics: Epic<T, S>[]): Epic<T, S>;
// new overload
declare function combineEpics<T>(...epics: T[]): T;
```

```ts
interface CustomEpic {
  (action$: ActionsObservable<Action>, store: MiddlewareAPI<State>, api: SomeApi): Observable<T>;
}

const epic1: CustomEpic = (action$, store, api: SomeApi) =>
  action$.ofType('STUFF').etc();

const rootEpic: Epic<Action, State> = (action$, store) =>
  combineEpics<CustomEpic>(epic1)(action$, store, api);
```

inuscript:
@jayphelps
Thank you for good advice. I change to this.

jayphelps:
Seems good to me. Wonder if @kwonoj could spare a second to confirm this solution is kosher?

index.d.ts
```diff
@@ -45,5 +45,7 @@ interface Options {
 }
 
 export declare function createEpicMiddleware<T, S>(rootEpic: Epic<T, S>, options?: Options): EpicMiddleware<T, S>;
+export declare function createEpicMiddleware<E, T, S>(rootEpic: E, options?: Options): EpicMiddleware<T, S>;
```
jayphelps: Is this overload necessary? Shouldn't the rootEpic always be `(action$, store)` only?

inuscript: Thank you. I didn't know this.
https://github.com/redux-observable/redux-observable/blob/master/src/createEpicMiddleware.js#L34

It's not good overload. I removed it.

kwonoj:
This seems bits complicated to me, so this is based on assumption custom epics are following `<E>` always, is that correct? Though it's general problem of generic that can't carry multiple types easily, if there's non-formed type of epic than it won't be accepted. Like, even default epic definition falls into those category like

```js
const defaultEpic = (a, s) => {};
const customEpic: CustomEpic<U> = (...) => {};

combineEpics<CustomEpic<U>>(defaultEpic, customEpic); //can't apply both epics?
```

jayphelps:
@kwonoj hmm I'm not 100% sure what you mean. This new overload is so that people can provide a custom signature for epics--usually because they're doing dependency injection. All of their epics  have to have the same signature.

kwonoj:
Yes, my initial assumption was someone would expect mixed types between different types of Epics. If this overload is intended to be used for constraint to epic have all same custom interfaces, I think it'll be fine.

jayphelps:
@kwonoj Someone could presumably provide a union of multiple signatures as well AFAICT

jayphelps:
Thanks @inuscript ! 🎉

PR #201, opened 2017-02-19
feat(build): es modules

Shyam-Chen:
:snowman:


package.json
```diff
@@ -77,6 +79,7 @@
     "babel-plugin-transform-object-rest-spread": "^6.8.0",
     "babel-polyfill": "^6.13.0",
     "babel-preset-es2015": "^6.13.2",
+    "babel-preset-latest": "^6.22.0",
```
jayphelps: Can we change to the specific plugins/presets required instead of just the latest?

Shyam-Chen: done

PR #209, opened 2017-03-06
Dependencies can now be `any` and is optional

jayphelps:
Also made the typings test more strict, disallowing implicit any

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -43,7 +43,7 @@ interface Adapter {
 
 interface Options {
   adapter?: Adapter;
-  dependencies: { [key:string]: any }
+  dependencies?: { [key: string]: any } | any;
 }
 
 export declare function createEpicMiddleware<T, S>(rootEpic: Epic<T, S>, options?: Options): EpicMiddleware<T, S>;
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "clean": "rimraf lib temp dist",
     "check": "npm run lint && npm run test",
     "test": "npm run lint && npm run build && npm run build:tests && mocha temp && npm run test:typings",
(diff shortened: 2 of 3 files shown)


index.d.ts
```diff
@@ -43,7 +43,7 @@ interface Adapter {
 
 interface Options {
   adapter?: Adapter;
-  dependencies: { [key:string]: any }
+  dependencies?: { [key: string]: any } | any;
```
jayphelps: Ideally we should use generics, but this is tough because then we'd need to have have conflicting overloads for `createEpicMiddleware`

```ts
declare function createEpicMiddleware<Action, State>(rootEpic: Epic<Action, State>, options?: Options<any>): EpicMiddleware<Action, State>;
declare function createEpicMiddleware<Action, State, Dependencies>(rootEpic: Epic<Action, State, Dependencies>, options?: Options<Dependencies>): EpicMiddleware<Action, State>;
```

Which would require an overload for the `Epic` interface, but you cannot have overloads with differing number of generic params

```ts
declare interface Epic<Action, State> {
  (action$: ActionsObservable<Action>, store: MiddlewareAPI<State>): Observable<Action>;
}

// not possible
declare interface Epic<Action, State, Dependencies> {
  (action$: ActionsObservable<Action>, store: MiddlewareAPI<State>, dependencies: Dependencies): Observable<Action>;
}
```

In TS 2.3 we'll be able to have default generic parameters, which will solve this I think:

```ts
declare interface Epic<Action, State, Dependencies = any> {
  (action$: ActionsObservable<Action>, store: MiddlewareAPI<State>): Observable<Action>;
  (action$: ActionsObservable<Action>, store: MiddlewareAPI<State>, dependencies: Dependencies): Observable<Action>;
}
```

PR #223, opened 2017-03-21
Proposal: Add a transformEpic helper

philpl:
I recently ran into a minor problem and came up with a small solution. I thought I should just send a PR, since I already have the code, and it should be easier to understand this way :smile:

It's a `transformEpic` helper that takes an epic changes it with a transformer that is an epic itself. The signature is thus:

```ts
transformEpic(epic: Epic<T, S>, transformer: Epic<T, S): Epic<T, S>
```

In a project I ran into a problem, where I need to catch all errors on every epic to have a fallback. Thus I want to `combineEpics` to create a root epic, then create another epic, that wraps around the root epic, executes it, and calls `catch` on action$.

```ts
transformEpic(
  rootEpic,
  action$ => action$::_catch(err =>of({ type: 'SET_ERROR_STATE', payload: err })
)
```

This might be nicer than doing this, especially in TypeScript:

```
(...args) => rootEpic(...args)::_catch(err =>of({ type: 'SET_ERROR_STATE', payload: err })
```

I realise this is really just about semantics, but finally I'm able to find an excuse to contribute to redux-observable :laughing:

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -50,3 +50,5 @@ export declare function createEpicMiddleware<T, S>(rootEpic: Epic<T, S>, options
 
 export declare function combineEpics<T, S>(...epics: Epic<T, S>[]): Epic<T, S>;
 export declare function combineEpics<E>(...epics: E[]): E;
+
+export declare function transformEpic<T, S>(epic: Epic<T, S>, transformer: Epic<T, S>): Epic<T, S>;
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,5 @@
 export { createEpicMiddleware } from './createEpicMiddleware';
 export { ActionsObservable } from './ActionsObservable';
 export { combineEpics } from './combineEpics';
+export { transformEpic } from './transformEpic';
 export { EPIC_END } from './EPIC_END';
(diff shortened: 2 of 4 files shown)


jayphelps:
This is indeed interesting. I haven't seen anyone need something similar to this so I'm hesitant to add it by default, but there may be some nice use cases we can create to justify it. I'll think about this some more, do you have any additional ones?

Don't worry about editing it just yet, but what do you think about `wrapEpic` instead? It seems like that more accurately reflects what it's doing--in fact, it's really taking two epics as arguments and returning a third, wrapped epic. In that case, we might as well make it generic enough to support an arbitrary number of epics, probably wrapping from right to left (like this does). It would then be sort of like `Object.assign` for epics.

philpl:
@jayphelps I think there are a couple, like modifying every action of a group of epics, starting more sideffects for a group of epics, adding analytics / meta data, globally transforming the stream of actions (error escape hatch).

But I think it's easy to write this pattern yourself once you need it in a project. So this might be useful to have, so people realise that they can do this. It might also be enough to add something to the documentation as an example on how to do this?

philpl:
Also +1 on the arbitrary number of epics. That would be awesome to have as well, since that adds more value and replaces combine epics where it makes sense?

BerkeleyTrue:
@jayphelps and @philpl I'd be down for a composeEpics (like the one proposed in #226) that can take an arbitrary number of epics. That would add a lot of value and cover this use case as well.

BerkeleyTrue:
@philpl Are you still interested in this PR?

Issue #204, opened 2017-02-22
suggestion for testing epics


continuata:
## Testing Epics

I've been looking at various ways of testing epics, and the following method is very simple and working great for our dev team.

I have created a `testEpic.js` module that takes an epic, an action, the number of expected actions and an optional state and returns an array with the resulting actions in a callback. We can then do a deep equals between the expected actions and the real actions. Those two arrays can also be sorted before comparison if the order of actions returned is unpredictable. Mocks and stubs can be cancelled and checked too in the callback.

### Epic test-harness
```es6
\\ testEpic.js
import { ActionsObservable } from 'redux-observable'
import { Subject } from 'rxjs'

export default (epic, count, action, callback, state = {}) => {
  const actions = new Subject()
  const actions$ = new ActionsObservable(actions)
  const store = { getState: () => state }
  epic(actions$, store)
    .take(count)
    .toArray()
    .subscribe(callback)
  if (action.length) {
    action.map(act => actions.next(act))
  } else {
    actions.next(action)
  }
}
```

### Example Epic
```es6
\\ fetchFooEpic.js
import { Observable } from 'rxjs'

export default (actions$) =>
  actions$.ofType('ADD_ERROR')
    .switchMap(() => Observable.merge(
      Observable.of({ type: 'CANCEL_ERROR' }),
      Observable.of({ type: 'ERROR_FULFILLED' })
     ))
```

### Example test
```es6
import mocha from 'mocha'
const { it, describe } = mocha
import { assert } from 'chai'
import fetchFooEpic from './fetchFooEpic'
import testEpic from './testEpic'

mocha.setup('bdd')

describe('fetchFooEpic', () => {
  it('calls the correct API', (done) => {
    const action = { type: 'ADD_ERROR', value: 'Test' }
    const expectedActions = [{ type: 'CANCEL_ERROR' },
                             { type: 'ERROR_FULFILLED' }]

    testEpic(fetchFooEpic, expectedActions.length, action, (actions) => {
      assert.deepEqual(actions, expectedActions)
      done()
    })    
  })
})

mocha.run()
```

jayphelps:
Looks pretty good. thank you much for sharing! I'm collecting these suggestions to better improve the recommendations in the docs.

btw, you might be happy to know that `Observable.of` supports an arbitrary number of arguments.

```js
// instead of this
Observable.merge(
  Observable.of({ type: 'CANCEL_ERROR' }),
[...]

PR #225, opened 2017-03-25
Add tests for epics to example app

### LINKED ISSUE #21: Add tests to the example app
Any plans to add tests to the example?

etiennedi:
Building on top of the work of  #204 (test helper) and #193 (dependency injection), it has become quite easy to write unit tests. However, the example app ("navigation") is still missing tests.

This PR
- adds tests for all epics in the example app (100% coverage on epic files)
- uses jest as the test runner
- adheres to the liniting styles of the main repo
- does not add tests for the rest of the app (reducers, components, containers, ...)

To see the test output run `npm t` or `npm t -- --verbose --coverage` for a more detailed output.

**Up for discussion**:
I'm not entirely satisfied with the way I'm testing `adminAccess.js`. This epic has a delay of 2000ms and a timer of 2000ms. As a result the test takes more than 4 seconds - most of which is spent doing nothing but waiting. I'm relatively new to RxJS (and reactive programming in general). Thus I'm not too sure what the best way to test such delays and timers is. One possible way would be to stub `Observable.prototype.delay` to something that completes after only a couple of milliseconds, but I'm not sure if this is the right approach. In this case it wouldn't alter the outcome, but in others it might.

**Docs**:
If you agree with my testing style and this is merged, I'd be happy to create another PR to update the docs. When I first encountered redux-observable, I thought I could never use it in production because of the line "Testing async code that creates side effects isn't easy. We're still learning the best way to test Epics." With DI and the suggested test helper from #204, I think testing epics has become quite easy. I'd feel comfortable suggesting the way used in this PR as a the standard way of testing epics in the docs. Hope you agree.

The epicTestHelper should of course be released as a separate npm package, so others won't have to copy/paste the file into their own code. 

Closes #21 
---


jayphelps:
This is super awesome!

I've been pretty hesitant to add more testing examples and update the docs because testing IRL still isn't perfect--the issues are around testing _time_.

```js
const somethingEpic = action$ =>
  action$.ofType(SOMETHING)
    .delay(1000)
    .mergeMap(() =>
      doSomething()
    );
```

How do I reliably test that this epic delays the side effect by 1000ms? Not easily, currently. RxJS has `TestScheduler` with marble diagrams, but it's not yet productized for common usages (e.g. arbitrary, especially lengthy, time ranges hard to notate).

We have lots of plans and discussions around fixing `TestScheduler`, but so far nothing has been shipped. Until then, it's kind of an embarrassing sore that most people don't realize.

***

Regarding this PR in specifically, I'm also hesitant to demonstrate helper utilities like `testEpic` because people tend to take repo example as gospel. I'm not outright saying no! 😄 I'm just hesitant.

Let me think some more, but I'd appreciate more discourse on this from you and anyone else reading this. Opinions are super valuable.

kockok:
Please also upgrade react-router to v4+. I'm struggling to resolve this problem.
screen shot 2017-04-16 at 6 59 36 pm

evertbouw:

PR #254, opened 2017-05-30
Allow using startWith/from/of inside epics.

### LINKED ISSUE #253: Actions emited with Observable.{startWith,of,from} are swallowed silently.
A bug.

Actions emited from epics using `startWith`, `of` or `from` are not processed by epics.
No error message is printed to developer console.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsbin.com or similar.**
https://jsbin.com/zanuvuxape/1/edit?html,css,js,output
```
const fetchUserEpic = action$ =>
  action$.ofType(FETCH_USER)
    .mergeMap(action => fakeAjax(`/api/users/${action.payload}`))
    .map(response => fetchUserFulfilled(response))
    .startWith(fetchUser('redux-observable'))
    // BUG: This action reaches reducers but not epics
```

Actions emitted this way should be processed by epic.
A hacky workaround is to add .delay(0) at the end. This is difficult to come up with for new users since there is no error message.

Observed on both Chrome and Firefox using redux-observable version 0.14.1.

This can be fixed by adding `.observeOn(Rx.Scheduler.asap)` in createEpicMiddleware
before `.subscribe(store.dispatch)`.

franekp:
Closes #253

- [+] I have read the [Contributor Guide](https://github.com/redux-observable/redux-observable/blob/master/CONTRIBUTING.md)
- [+] My commit messages are in [conventional-changelog-standard](https://github.com/redux-observable/redux-observable/blob/master/CONTRIBUTING.md#sending-a-pull-request) format.

### DIFF
--- a/src/createEpicMiddleware.js
+++ b/src/createEpicMiddleware.js
@@ -1,6 +1,11 @@
 import { Subject } from 'rxjs/Subject';
+import { ReplaySubject } from 'rxjs/ReplaySubject';
 import { map } from 'rxjs/operator/map';
 import { switchMap } from 'rxjs/operator/switchMap';
+import { mergeAll } from 'rxjs/operator/mergeAll';
+import { window } from 'rxjs/operator/window';
+import { startWith } from 'rxjs/operator/startWith';
+import { zipStatic } from 'rxjs/operator/zip';
 import { ActionsObservable } from './ActionsObservable';
 import { EPIC_END } from './EPIC_END';
 
@@ -11,6 +16,52 @@ const defaultAdapter = {
 
 const defaultOptions = {
   adapter: defaultAdapter
+};
+
+class SluiceGate {
+  static createClosed() {
+    return new this(false);
+  }
+  static createOpen() {
+    return new this(true);
+  }
+  constructor(isOpen) {
+    if (typeof isOpen !== 'boolean') {
+      throw new TypeError('SluiceGate constructor expects one boolean argument.');
+    }
+    this._is_open = isOpen;
+    this._open = new Subject();
+    this._close = new Subject();
+  }
+  sluice(input) {
+    return zipStatic(
+      input::window(this._close)::map(
+        (w) => {
+          const rep = new ReplaySubject();
+          w.subscribe(rep);
+          return rep;
+        }
+      ),
+      (this._is_open ? this._open::startWith({}) : this._open),
+      (w, _) => w,
+    )::mergeAll();
+  }
+  open() {
+    if (!this._is_open) {
+      this._open.next({});
+      this._is_open = true;
+    }
+  }
+  close() {
+    if (this._is_open) {
+      this._close.next({});
+      this._is_open = false;
+    }
+  }
+}
+
+const sluice = function (gate) {
+  return gate.sluice(this);
 };
 
 export function createEpicMiddleware(epic, options = defaultOptions) {
@@ -26,6 +77,7 @@ export function createEpicMiddleware(epic, options = defaultOptions) {
     new ActionsObservable(input$)
   );
   const epic$ = new Subject();
+  const outputGate = SluiceGate.createClosed();
   let store;
 
   const epicMiddleware = _store => {
@@ -45,10 +97,14 @@ export function createEpicMiddleware(epic, options = defaultOptions) {
           return output$;
         })
         ::switchMap(output$ => options.adapter.output(output$))
+        ::sluice(outputGate)
         .subscribe(store.dispatch);
 
       // Setup initial root epic
       epic$.next(epic);
+      // wait with any calls to store.dispatch until middleware
+      // initialization is complete
+      setTimeout(() => outputGate.open());
 
       return action => {
         const result = next(action);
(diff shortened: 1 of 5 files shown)


src/createEpicMiddleware.js
```diff
@@ -45,6 +48,7 @@ export function createEpicMiddleware(epic, options = defaultOptions) {
           return output$;
         })
         ::switchMap(output$ => options.adapter.output(output$))
+        ::mergeMap(action => of(action, asap))
```
jayphelps: hmm I'm pretty reluctant to schedule every action asap. Seems like it could create hard to understand behavior for people, as well as more complex stack traces--separately, I think `::observeOn(asap)` would be the more idiomatic way of achieving asap scheduling, though they both would have the same effect seemingly. Don't worry about changing it just yet, need to figure out out if this is the best solution in general.

jayphelps:
Will need to think about this one some more, but my gut thinks the solution is to wait until after the middleware is finished setting up before [subscribing to the root epic](https://github.com/redux-observable/redux-observable/blob/master/src/createEpicMiddleware.js#L51). That's tricky because of the way the redux middleware API is designed as well as we want to be careful not to miss any actions dispatched synchronously outside of redux-observable (e.g. in server rendering cases this is common)

I'm pretty reluctant to schedule _all_ actions on asap scheduler cause that feels like it might cause  other problems and make the normal flow harder to reason for people--it also wouldn't be as performant but I'm not as worried about that.

test/createEpicMiddleware-spec.js
```diff
@@ -11,6 +11,13 @@ import { of } from 'rxjs/observable/of';
 import { empty } from 'rxjs/observable/empty';
 import { mergeStatic } from 'rxjs/operator/merge';
 import { mapTo } from 'rxjs/operator/mapTo';
+import { startWith } from 'rxjs/operator/startWith';
+
+const delay = async () => {
+  for (const _ of new Array(10)) {  // eslint-disable-line no-unused-vars
+    await (new Promise(setTimeout));
+  }
+};
```
BerkeleyTrue: Why not use Observables to declaratively add these delays?

src/createEpicMiddleware.js
franekp: I tried using `::observeOn(asap)` first, but it broke the `it('should throw if you provide a root epic that doesn\'t return anything)'` since exception was thrown from the `asap` scheduler and not from `createStore(..., applyMiddleware(...))` call.

jayphelps: oh right cause observeOn will schedule _everything_ whereas mergeMap+of(a, asap) will just schedule next, not error/complete

test/createEpicMiddleware-spec.js
franekp: Sure, I will add change it.

franekp:
>the solution is to somehow wait until after the middleware is finished setting up before subscribing to the root epic.

Hmm, it turns out that there 2 problems here:
1. Actions dispatched during middleware initialization do not reach middleware
    https://github.com/reactjs/redux/issues/1240 - seems to not going to be fixed on redux side
2. "race condition" between `epic$.next(...)` and synchronous actions emitted by epic:
    my guess is that somehow these actions are emitted before `epic$.next(...)` wires
    `action$` to the input of new epic.

Waiting until after middleware is initialized solves (1) but not (2).

BerkeleyTrue:
@franekp I'd wait to make any changes to this PR.

The most likely scenario is we move to a store enhance in place of a middleware. This will give us the flexibility to manage some edge cases that have popped up as well as add some power features not available to use using middleware API.

See: https://github.com/redux-observable/redux-observable/issues/256

franekp:
> I'd wait to make any changes
Hmm, too late.

@jayphelps Fixed without changing schedulers now.

mpiroc:
Any update on this? I think it may be the root cause of a tricky bug I'm struggling with in my app.

jayphelps:
@mpiroc Should be fairly easy to verify--check if you are emitting an action immediately on app start via startWith or concat, and another epic is listening for this action but not receiving it but your reducers are. This is a fairly odd pattern IMO, but I don't doubt there are some rare use cases for it.

We're switching from a middleware to a store enhancer in #256. This is progressing, but admittedly slow because this is all volunteer work.

mpiroc:
Thanks @jayphelps, my bug turned out to be unrelated to this issue.

jayphelps:
@franekp I feel pretty bad that that is a pretty impressive PR, but has sat unmerged for so long. I'm sorry! It's still a use case I think we should support, I just don't have confidence yet this approach is the way to achieve it. I'm going to try to figure out #256 in the coming weeks, then we can circle back on this.

franekp:
No problem, I understand. I am glad you want to support this use case.

jayphelps:
I believe this was fixed last year. Definitely lmk if not! https://github.com/redux-observable/redux-observable/commit/62a8c7390ebd4c59e38e52e19bb2e05ccdff601c

Issue #256, opened 2017-06-01
RFC: add createEpicEnhancer, deprecate createEpicMiddleware


BerkeleyTrue:
There are currently two edge cases to Redux-Observable (RO) that don't have a clear an easy solution:

* [Action order discrepancy](https://github.com/redux-observable/redux-observable/issues/252)
* [Sync actions emitted at subscription are not seen by epics](https://github.com/redux-observable/redux-observable/issues/253)

While both of these issues can be worked around some Rx know-how and grit, it will catch seasoned people by surprise (I have been caught up by the second issue). 

The major problem underlying both of these issues is the middleware API. It limits how we can intercept and dispatch actions going to and coming from epics. While there are possible solutions to both of these problems that would keep using the middleware API, they have their own set of tradeoffs and breaking changes. 

I suggest we consider adding a createEpicEnhancer function that will create a store enhancer. This should give us the flexibility to cover these two issues (may help simplify SSR as well). At the same time we should deprecate createEpicMiddleware. 

This would be a breaking change and should be thoroughly discussed by the community.

jayphelps:
I'm 100% on board with the idea if someone has the cycles to do it. #100 aimed to do this, but has been stagnant for a while. It would also allow us to truly decide whether we want to do #56

BerkeleyTrue:
@jayphelps I'd definitely want to invest time in this. There are a lot of hacky things I'm doing to get SSR working that could be made easier with an enhanced store vs middleware.

PR #259, opened 2017-06-08
feat: Add createEpicEnhancer API

### LINKED ISSUE #252: Action order discrepancy (reducers vs epics)
Bug

In certain situations, the order of the actions that are received by the epics is not the same that is received by the reducers (and store subscribers).

Example: https://jsbin.com/fuzoba/1/edit?html,js,console (if it doesn't run, switching the ES6/Babel tab to Javascript and then back to ES6/Babel fixes it https://github.com/jsbin/jsbin/issues/2668)

Although it is a contrived example, one can see that the intended behaviour was to cancel **PING** actions due to some business logic within a store subscriber, and as such the expected flow should be limited to: 

> PING -> CANCEL

Instead, the pingEpic receives the **CANCEL** action before the **PING** action, which eventually leads to a **PONG** action to be dispatched, since it was not canceled correctly:

> reducers: PING -> CANCEL -> PONG
> epics: CANCEL -> PING -> PONG

As far as I understand, the order of the actions received by Epics should be the same as the one received by the store (reducers). Since epics are often used as process managers, the order of the actions is quite important, specially when using **race** or **takeUntil** based operations.

I first detected this issue in version 0.12.1 and then again, after upgrading, in version 0.14.1. 

After digging into the source code I realized that the problem probably lies on the main block within **createEpicMiddleware** (with additional comments):
```
return action => {
  //action incoming to epicMiddleware (after any higher priority middleware) 
  const result = next(action);
  //action will be dispatched to epics (after any lower priority middleware, reducers, subscribers)
  input$.next(action);
  //done
  return result;
};
```

The problem is that `const result = next(action);` can potentially lead to other actions being synchronously dispatched (by other middleware or subscribers) triggering a new "store update flow". That means that actions' scheduling for epics' input can behave as a stack (FILO) instead of a queue (FIFO).

One naive solution could involve queuing actions explicitly, something like the following:
```
...
const epic$ = new Subject();
let store;
const pendingActions = [];
...
return action => {
  pendingActions.push(action);
  const result = next(action);
  const nextAction = pendingActions.shift();
  input$.next(nextAction);
  return result;
};
```

What do you think?

_PS: Special thanks to all redux-observable contributors and the community for your sweat, tears and coolness!_

BerkeleyTrue:
This PR introduces a new function `createEpicEnhancer` to eventually replace `createEpicMiddleware`. By moving from the redux middleware to store enhancer API, we the ability to control the dispatch function directly and fix some edge case issue.

This may make using [react-redux-epic](https://github.com/BerkeleyTrue/react-redux-epic) simpler by introducing some new api directly to the store to control the lifecycle of the `action$` observable.

TO DO:

- [x] Update queueUntilType to use functional approach
- [x] Add spec for queueUntilType operator
- [ ] Add documentation
- [ ] Update typescript definitions

closes #252, #253, #256


jayphelps:
What the usecase for `cacheUntilType`?

BerkeleyTrue:
@jayphelps The cacheUntilType operator caches actions untill the `EPIC_INIT` action has made it through the epics. This fixes #253

src/cacheUntilType.js
```diff
@@ -0,0 +1,51 @@
+import { OuterSubscriber } from 'rxjs/OuterSubscriber';
+import { subscribeToResult } from 'rxjs/util/subscribeToResult';
+
+
+export function cacheUntilType(type) {
+  return this.lift(new CacheUntilTypeOperator(type));
+}
+
+class CacheUntilTypeOperator {
```
jayphelps: don't do anything yet, but if we do go this route I'd rather use closures instead of classes so that we have a smaller bundle size since this is should not be a hot code path.

BerkeleyTrue: Sure. Went with classes original as it seemed to be the pattern established by RxJSv5 source but I'm not attached.

jayphelps: totally. FYI it does that for perf reasons cause _at the time_ closures were much slower than objects in v8. Since then they've released turbofan and shit...so it may no longer hold true. :man_shrugging:

jayphelps: I _think_ this is an equivalent closure version, probably even smaller if you compared the compiled output of the class version. Don't need to use my logic, just providing for future reference in case it's not clear what I meant. (this isn't a pattern really documented anywhere)

```js
export function queueUntilType(expectedType) {
  return this.lift(function (source) {
    let queue = [];
    let hasSeenType = false;

    return source.subscribe({
      next: action => {
        // It's important to check this first because of possible re-entry
        // before we've finished flushing the queue.
        if (hasSeenType) {
          this.next(action);
          return;
        }

        if (action && action.type === expectedType) {
          hasSeenType = true;
          this.add(subscribeToResult(this, queue));
          // no longer needed, free memory
          queue = null;
        } else {
          queue.push(action);
        }
      },
      error: e => this.error(e),
      complete: () => this.complete()
    });
  });
}
```

***

I did the conditional differently but I believe it has the same effect. 😝  Your way is totally fine.

BerkeleyTrue: Didn't realize you could pass a function to `lift` but now that I see it makes sense. The operator is always called using the `call` method and bound to the subscriber. Neat!

jayphelps: @BerkeleyTrue yeah, that was the intent all along, however it's a little "clever" and a bit contentious to some.

jayphelps:
This will needs docs too before merge.

What do you think the deprecation cycle for `createEpicMiddleware` should look like? I think we might want to introduce the enhancer and let it sit without deprecated the middleware for a month or so.

BerkeleyTrue:
@jayphelps Agreed we should take it slow.

1. Add middleware as a feature release.
2. Wait a couple of months and release a major release that adds the deprecation warning
3. Another couple of months and release another major removing epicMiddleware.

What do you think about making this final release the 1.0.0 release?

jayphelps:
@BerkeleyTrue sounds like a good thought to make it 1.0 after the deprecate then remove. Ultimately I've been hesitant to bump 1.0 simply because I want the testing story to be solid and _documented_ tho.

BerkeleyTrue:
@jayphelps That could also be part of the release now that we have a solid plan for the future of the app.

BerkeleyTrue:
@jayphelps Updated to use functional style operator

src/queueUntilType.js
```diff
@@ -0,0 +1,35 @@
...
+          this.add(subscribeToResult(outer, queue));
+          // no longer needed, free up memory
+          nullifyQueue();
+        } else {
+          queue.push(action);
+        }
+      },
+      error: e => this.error(e),
+      complete: () => this.complete(),
+      unsubscribe: nullifyQueue
```
jayphelps: AFAIK passing `unsubscribe` isn't a supported thing for subscribers. I don't think that would be a leak anyway because if it's unsubscribed, presumably the containing closure would be collected (unless something else was leaking it)

BerkeleyTrue: > AFAIK passing unsubscribe isn't a supported thing for subscribers

As in this is something that should not be done?

From tracing the source code I can see it the observer is eventually wrapped in a SafeSubscriber, which will parse out the unsubscribe method correctly. http://reactivex.io/rxjs/file/es6/Subscriber.js.html#lineNumber160

> presumably the containing closure would be collected

I can remove it in this case.

jayphelps: it was just AFAIK, so AFAIK passing unsubscribe like that does nothing at all. I think the object will get passed to `new Subscriber(yourObserver)` which will internally wrap it in `SafeSubscriber` neither of which (again AFAICT) call unsubscribe on any passed in observer object. Quick testing suggests this as well: https://jsbin.com/dozuseg/edit?js,console

Certainly I _could be wrong_ as I have many times hehe 😈 If I am, this would be very interesting to know so lmk if you can reproduce.

jayphelps: Oh I missed your linking to the source. Hmm that does indeed suggest at first glance that it would be called..but I can't reproduce. This is obv secondary to this PR, I'm just indeed very curious! 😄

jayphelps: we're digging into this btw.. some interesting examples so far:

* called on error https://jsbin.com/zudivez/edit?js,console (edited)
* called on complete https://jsbin.com/nazixix/edit?js,console (edited)
* NOT called on unsubscribe https://jsbin.com/lerilub/edit?js,console (edited)

(but `finally()` is called for all three)

Current thinking is either this shouldn't be supported at all or should have the same behavior as `finally()` and thus is a bug. Not sure which yet.

BerkeleyTrue: > I'm just indeed very curious!

Yes now I am curious as well. My intuition tells me this is somehow a consequence of the lift pattern. That the unsubscribe logic must be bundled with the observer (or what is now internally called the subscriber).

Yes the unsubscribe is called in the above code:

hyper

hyper

It is strange that in your finally example code above it is not called. I believe that is most likely a bug.

BerkeleyTrue: I'm unable to reproduce reliably either. This is pretty frustrating.

https://jsbin.com/jirutayuru/edit?js,console

jayphelps: FYI I think that bug is going to be tracked here #2675

niamleeson:
Is the rxjs bug blocking the progress of allowing the usage of redux-observable as an enhancer?

BerkeleyTrue:
@niamleeson the actual code is ready an working. What's left is writing documentation and typescript definitions. Unfortunately I haven't had much time lately.

@jayphelps what do you think of merging this in and opening up the those two points to the community?

niamleeson:
@BerkeleyTrue I can help.

BerkeleyTrue:
@niamleeson Would you like to contribute documentation to this?

@jayphelps What do you think of my above comment?

jayphelps:
@BerkeleyTrue I'm against merging without docs. If we merged it, I wouldn't want to release it until it was documented and deprecations were added, so it being merged wouldn't be meaningful as-is. 

I've just seen too many times where it just leads to things being undocumented for long periods of time. e.g. rxjs is heavily bitten by this.

BerkeleyTrue:
Alright, I could give access to my personal fork to those who would like to help with documentation. I won't have time to work on this again till October otherwise

jayphelps:
@BerkeleyTrue someone can also make it as a separate PR here since they don't have any actual cross dependencies AFAIK?

BerkeleyTrue:
That's true. If anyone's open to it let us know here. I'll open up an issue as well.

niamleeson:
@BerkeleyTrue Yes I would like to help.

PR #274, opened 2017-07-09
docs:add Chinese localization

moreJs:
Lately I research the redux-observable and I find that it lack the Chinese Docs.
As we all know, the Chinese Docs lead the rxjs community to be stronger than ever.
So I add Chinese localization for the redux-observable . Hope this can help in anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,8 @@
 [Redux](http://github.com/reactjs/redux). Compose and cancel async actions to create side effects and more.
 
 [https://redux-observable.js.org](https://redux-observable.js.org)
-[https://redux-observable.js.org 中文版](https://redux-observable-cn.js.org/)
+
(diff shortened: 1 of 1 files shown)


jayphelps:
Wow this is freaking awesome!! If we're going to keep it hosted in a separate repo, I'd want our external linking here to say that it's an unofficial version since we cannot control what you say in it--e.g.  you could have Code of Conduct violations we wouldn't know about.

I don't know Chinese but here's what I mean in English

> https://redux-observable-cn.js.org Chinese version (unofficial)

Note that it's also important that the URL is the -cn one, not ours (otherwise it's misleading)

moreJs:
@jayphelps  I totally agree with you. so i add 'unofficial label'.

jayphelps:
Huge thank you thank you thank you! ❤️

Issue #102, opened 2016-09-06
RFC: onErrorReturn() operator


jayphelps:
Some Rx implementations contain an `onErrorReturn` operator. http://reactivex.io/documentation/operators/catch.html

> instructs an Observable to emit a particular item when it encounters an error, and then terminate normally

![image](https://cloud.githubusercontent.com/assets/762949/18284626/8e5cda00-741f-11e6-8a59-346e8275203b.png)

> The `onErrorReturn` method returns an Observable that mirrors the behavior of the source Observable, unless that Observable invokes onError in which case, rather than propagating that error to the observer, onErrorReturn will instead emit a specified item and invoke the observer’s onCompleted method. [javadoc](http://reactivex.io/RxJava/javadoc/rx/Observable.html#onErrorReturn(rx.functions.Func1))

This operator is usually similar to `.catch()` but that it expects your provided selector to return a single _value_ instead of an Observable.

In practice this is exactly what people do 99% of the time in redux-observable.

#### Using .catch()
```js
const fetchUserEpic = action$ =>
  action$.ofType(FETCH_USER)
    .mergeMap(action =>
      ajax.getJSON(`/api/users/${action.payload}`)
        .map(fetchUserFulfilled)
        .catch(error => Observable.of({
            type: FETCH_USER_REJECTED,
            payload: error.xhr.response,
            error: true
        }))
    );
```

#### Using .onErrorReturn()
```js
const fetchUserEpic = action$ =>
  action$.ofType(FETCH_USER)
    .mergeMap(action =>
      ajax.getJSON(`/api/users/${action.payload}`)
        .map(fetchUserFulfilled)
        .onErrorReturn(error => ({
            type: FETCH_USER_REJECTED,
            payload: error.xhr.response,
            error: true
        })
    );
```

I've noticed most people don't fully understand what `.catch()` is doing--that you're returning an observable that should be switched to, on error. `onErrorReturn` on the other hand has more obvious semantics.

This probably belongs in RxJS v5 core, but before I pitch it there I wanted to confirm the community agrees this would be super helpful. Worst case, we can add `onErrorReturn` to the `ActionsObservable` prototype, but I'm confident core will want it too.

BerkeleyTrue:
This is 99.99% of my use of the catch operator.

jayphelps:
After discussing with numerous people, I think this is a great addition to core so I've made a PR for this. We'll bikeshed there. https://github.com/ReactiveX/rxjs/pull/1928

[...]

PR #281, opened 2017-07-17
feat(ActionsObservable): catchAndRelease operator

rgbkrk:
Catches errors, releases an action with the error attached, and continues the stream.

Happy ~~fishing~~ observing!

🎣

In all seriousness, I've done this enough time in apps that I wish this was available, especially when helping people getting started with redux-observable.

### DIFF
--- a/src/ActionsObservable.js
+++ b/src/ActionsObservable.js
@@ -1,7 +1,9 @@
-import { Observable } from 'rxjs/Observable';
-import { of } from 'rxjs/observable/of';
-import { from } from 'rxjs/observable/from';
-import { filter } from 'rxjs/operator/filter';
+import { Observable } from "rxjs/Observable";
+import { of } from "rxjs/observable/of";
+import { merge } from "rxjs/observable/merge";
+import { from } from "rxjs/observable/from";
+import { filter } from "rxjs/operator/filter";
+import { _catch } from "rxjs/operator/catch";
 
 export class ActionsObservable extends Observable {
   static of(...actions) {
@@ -21,6 +23,14 @@ export class ActionsObservable extends Observable {
     const observable = new ActionsObservable(this);
     observable.operator = operator;
     return observable;
+  }
+
+  catchAndRelease(action) {
+    return this::_catch((error, source) => {
+      const fullAction = Object.assign({}, action, { error });
+
+      return merge(of(fullAction), source);
+    });
   }
 
   ofType(...keys) {


src/ActionsObservable.js
```diff
@@ -23,6 +25,17 @@ export class ActionsObservable extends Observable {
     return observable;
   }
 
+  catchAndRelease(action) {
+    return this::catch((error, source) => {
+      const fullAction = Object.assign({}, action { error });
```
jayphelps: `action { error }` looks like a typo missing a comma?

jayphelps:
It's actually not 100% clear how this is used. Can you describe?

We'd need documentation and tests before I'd merge this as well but this is a great way to start the conversation!

src/ActionsObservable.js
```diff
@@ -1,7 +1,9 @@
-import { Observable } from 'rxjs/Observable';
-import { of } from 'rxjs/observable/of';
-import { from } from 'rxjs/observable/from';
-import { filter } from 'rxjs/operator/filter';
+import { Observable } from "rxjs/Observable";
+import { of } from "rxjs/observable/of";
+import { merge } from "rxjs/observable/merge";
+import { from } from "rxjs/observable/from";
+import { filter } from "rxjs/operator/filter";
+import { filter } from "rxjs/operator/catch";
```
jayphelps: typo, identifier should be `_catch`

src/ActionsObservable.js
```diff
@@ -23,6 +25,14 @@ export class ActionsObservable extends Observable {
     return observable;
   }
 
+  catchAndRelease(action) {
+    return this::catch((error, source) => {
```
jayphelps: should be `_catch` because `catch` is a reserved word

jayphelps: feel free to ignore this for now as we're still discussing the viability in general

rgbkrk:
Many times in `redux-observable`, the pattern I've been using has been to run:

```
action$.ofType(SOMETHING)
  .map(...)
  .catch((error, source) =>
    Observable.merge(
      Observable.of(somethingErroredAction(error)),
      source
    )
```

When it would be really nice to run something like this:

```
action$.ofType(SOMETHING)
  .map(...)
  .catchAndRelease(somethingErroredAction())
```

Thanks for starting this discussion, sorry for the radio silence on redux-observable. :smile:

BerkeleyTrue:
this: 

```js
action$.ofType(SOMETHING)
  .map(...)
  .catchAndRelease(somethingErroredAction())
```

Wouldn't work since the observable returned from `.map()` is not an instance of `ActionObservable`. 

The way it's currently written it would need to be:
```js
const { catchAndRelease } = ActionsObservable;

action$.ofType(SOMETHING)
  .map(...)
  ::catchAndRelease(somethingErroredAction())
```

There are also a couple of assumptions made, like adding an `error` property to an action and re-subscribing to the source observable that will not work in every use case for this type of operator.

rgbkrk:
> There are also a couple of assumptions made, like adding an error property to an action and re-subscribing to the source observable that will not work in every use case for this type of operator.

That's so true @BerkeleyTrue. I'm hoping there are some patterns we can provide the means to use within redux-observable for errors, even if we have a few different ways of approaching it.

jayphelps:
The Observable returned from map is actually indeed an ActionsObservable unless there is a bug. We use lift.

BerkeleyTrue:
@jayphelps Ah yes, if you keep the chain straight from the actions observable it does. I was not aware of that.

But if a user starts with another observable, such as a defer or mergeStatic, the initial source is the plain observable.

jayphelps:
@rgbkrk I think there is merit in this operator, but will need some tweaks to make it less opinionated IMO (and a different name 🐟 )

I'm thinking we can add a couple more operators like this and [`onErrorReturn`](#102), but to be good stewards and try to make it much more clear these are **custom operators** we require users to explicitly "install" them. This also is good for the bundle size obsessed.

Something like:

```js
import 'redux-observable/add/operator/ofType';
import 'redux-observable/add/operator/onErrorReturn';
import 'redux-observable/add/operator/catchAndRelease';
```

I do have a concern that this will be extra confusing for people that think this adds the operators to Observable when instead it only adds it to ActionsObservable. I'd rather not do the former ever in redux-observable itself.

Anyone have thoughts/opinions?

rgbkrk:
I'm definitely all for the `redux-observable/add/operator/` variety, as well as the change in name. I was being a bit silly, for lack of a better name.

> I do have a concern that this will be extra confusing for people that think this adds the operators to Observable when instead it only adds it to ActionsObservable. I'd rather not do the former ever in redux-observable itself.

Glad you cleared that up, because I didn't realize that would be the case. Could always make the path terrible like `import 'redux-observable/addToActionsObservable/operator/onErrorReturn';`

src/ActionsObservable.js
```diff
@@ -23,6 +25,14 @@ export class ActionsObservable extends Observable {
     return observable;
   }
 
+  catchAndRelease(action) {
+    return this::_catch((error, source) => {
+      const fullAction = Object.assign({}, action, { error });
```
jayphelps: How bout it's just agnostic of actions in general:

```js
onErrorReturnAndRetry(selector) {
  return this::_catch((error, source) => {
    const value = selector(error);
    return merge(of(value), source);
  });
}
```

```js
const somethingErrored = error => ({ type: SOMETHING_ERRORED, error });

action$.ofType(SOMETHING)
  .switchMap(...)
  .onErrorReturnAndRetry(error => somethingErrored(error))
```

Or does that make it too less useful compared to just doing the catch/of/merge yourself?

jayphelps:
> Glad you cleared that up, because I didn't realize that would be the case. Could always make the path terrible like import 'redux-observable/addToActionsObservable/operator/onErrorReturn';

Yeah, something more obvious in the name might help.

I was considering removing them entirely from redux-observable and instead putting them in a standalone library like https://github.com/cartant/rxjs-etc or our own; then just recommending them in the docs. But that doesn't work with `ofType`, which is pretty specific to redux since it checks the type property ☹️ 

Off the top of my head maybe the `type` property is common enough in JS to be an OK convention to support (e.g. ASTs commonly use it). `ofType` could have an overload: if you pass an object it does `value instanceof type`, if you pass anything else (like a string or symbol) it does `value.type === type`.

rgbkrk:
I'm going to close this as an inactive PR from me. Thanks for all the discussion!

PR #288, opened 2017-07-30
Add generic option for `ofType` (TypeScript).

philcockfield:
Adds a type definition allowing the `ActionsObservable.ofType` method to take a strongly typed action name.

## Example:

    const fooEpic = (action$: ActionsObservable<IMyAction>) =>
      action$
        .ofType<IMyAction['type']>('@module/LOAD')

## Rationale:
This allows renaming of strongly typed action string (in say VSCode), which will update all references to the "magic string" across a project.

PR #250, opened 2017-05-23
feat(typings): added dependencies generic type

rafaelkallis:
This PR is a *draft* and should be considered as an extension to the discussion started in #231.
Feedback is mandatory ;)

Added a dependency generic type to typings.
The type defaults to `any` for backwards compatibility.

- [x] I have read the [Contributor Guide](https://github.com/redux-observable/redux-observable/blob/master/CONTRIBUTING.md)
- [x] My commit messages are in [conventional-changelog-standard](https://github.com/redux-observable/redux-observable/blob/master/CONTRIBUTING.md#sending-a-pull-request) format.

PR #289, opened 2017-08-02
Fix typing of `ofType`

pelotom:
By making the type parameter `T` extend `Action` (i.e. telling the compiler that `T` has a `type` field),
it's possible to type `ofType()` correctly using lookup types.

- [x] ⚠️ My commit messages are in [conventional-changelog-standard](https://github.com/redux-observable/redux-observable/blob/master/CONTRIBUTING.md#sending-a-pull-request) format.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,9 +1,9 @@
-import { Middleware, MiddlewareAPI } from 'redux';
+import { Middleware, MiddlewareAPI, Action } from 'redux';
 import { Observable, ObservableInput } from 'rxjs/Observable';
 import { Scheduler } from 'rxjs/Scheduler';
 import { Operator } from 'rxjs/Operator';
 
-export declare class ActionsObservable<T> extends Observable<T> {
+export declare class ActionsObservable<T extends Action> extends Observable<T> {
   /**
    * Just like RxJS itself, we can't actually make this method always type-safe
    * because we would need non-final position spread params e.g.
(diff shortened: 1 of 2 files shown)


jayphelps:
Thanks @pelotom for the PR.

I'm worried it will prevent people from using things that aren't actually actions, like dispatching thunks or other objects that don't have the Action shape--because they're using some other middleware as well. I've seen this in the wild and think it should be supported. Thoughts?

pelotom:
My feeling is that I'd rather have a conservative default assumption of what you're dealing with (actions), and if the user knows they might also be dealing with something else they can override this assumption by making an explicit type cast. As opposed to assuming nothing whatsoever and having the default not be type safe. I also think that the majority of the time people are not using any other middleware if they're using `redux-observable`, and this common case should be what the typings attempt to capture and make safe.

jayphelps:
Hmm I'm not following? The current usage of generics should provide complete type safety if you set everything up correctly and provide the right type param--if this is not the case, it's a bug in our type definitions.

pelotom:
@jayphelps how do you figure? The type parameter is not used in the current type signatures for `ofType`:

```ts
```

pelotom:
@jayphelps how do you figure? The type parameter is not used in the current type signatures for `ofType`:`

```ts
ofType(...key: string[]): ActionsObservable<T>;
ofType(...key: any[]): ActionsObservable<T>;
```

The `key` arguments can be any strings you want, and don't necessarily have to match the `type` field of any of the possible actions that could be dispatched. This is not type safe.

jayphelps:
Sounds like I didn't read this PR closely enough as I didn't notice the ofType T["type"] changes and stuff. Sorry about that. I'll try and fully grasp what you're proposing. 👍

jayphelps:
Sorry for the delay--I don't have a ton of OSS time lately as I'm still recovering from surgery. If I don't get back to this within the next week feel free to ping me.

index.d.ts
```diff
@@ -1,38 +1,38 @@
...
+  static from<T extends Action>(ish: ObservableInput<T>, scheduler?: Scheduler): ActionsObservable<T>;
+  static from<T extends Action, R extends Action>(ish: ArrayLike<T>, scheduler?: Scheduler): ActionsObservable<R>;
 
   constructor(input$: Observable<T>);
-  lift<R>(operator: Operator<T, R>): ActionsObservable<R>;
-  ofType(...key: string[]): ActionsObservable<T>;
-  ofType(...key: any[]): ActionsObservable<T>;
+  lift<R extends Action>(operator: Operator<T, R>): ActionsObservable<R>;
+  lift<R>(operator: Operator<T, R>): Observable<R>;
+  ofType(...key: T['type'][]): ActionsObservable<T>;
```
philcockfield: 👍

philcockfield:
Agreed @pelotom, until `ofType` can take a type generic, I need to stick with 

     action$.filter(a => a.type === 'FOO')

For type safety.  It's ugly-er...but the ability to safely rename actions, and other type-safety around the action types, is very desirable.

philcockfield:
I created a similar PR (#288) which does essentially the same thing, but you're seems to take into account a bunch of wider considerations than mine, so I'd say go with your.

index.d.ts
```diff
@@ -1,38 +1,38 @@
...
-  static from<T, R>(ish: ArrayLike<T>, scheduler?: Scheduler): ActionsObservable<R>;
+  static from<T extends Action>(ish: ObservableInput<T>, scheduler?: Scheduler): ActionsObservable<T>;
+  static from<T extends Action, R extends Action>(ish: ArrayLike<T>, scheduler?: Scheduler): ActionsObservable<R>;
 
   constructor(input$: Observable<T>);
-  lift<R>(operator: Operator<T, R>): ActionsObservable<R>;
-  ofType(...key: string[]): ActionsObservable<T>;
-  ofType(...key: any[]): ActionsObservable<T>;
+  lift<R extends Action>(operator: Operator<T, R>): ActionsObservable<R>;
+  lift<R>(operator: Operator<T, R>): Observable<R>;
```
jayphelps: This should still return and `ActionsObservable` (the same signature as it was before) or am I confused?

```ts
lift<R>(operator: Operator<T, R>): ActionsObservable<R>;
```

It's not super useful right now since it would only give you chaining `ofTypes()`, but later we may add additional operators besides `ofType`. e.g. #281

```ts
action$
  .ofType(SOMETHING)
  .debounceTime(200)
  .anotherCustomOperatorWeProvide()
```

jayphelps: hmmm thinking about this more, I think I realize why you changed it. It probably won't allow that signature because if the return type is `ActionsObservable<R>` R must now extend Action but `lift<R>` does not apply that restriction. i.e. ActionsObservable must always be a stream of actions...which makes sense by its _name_ but its intent was more general. Not sure what the best course is since this isn't a problem that exists today, but likely will at some point 😢

pelotom: Yeah, unfortunately we can't completely constrain the type parameter `R` of an inherited method. But by providing the 2 overloads

```ts
lift<R extends Action>(operator: Operator<T, R>): ActionsObservable<R>;
lift<R>(operator: Operator<T, R>): Observable<R>;
```

you get the next best thing: whenever you use an `Operator<T, R>` where `R extends Action`, it will assume the first signature, so the result type will be `ActionsObservable<R>`. 

In general, if you want to add methods to observables which _don't_ require the assumption that `T extends Action`, would it not work to augment the definition of `Observable` itself, similar to how RxJS itself adds its operators piecemeal:

```ts
declare module 'rxjs/Observable' {
    interface Observable<T> {
        myObservableOperator: Foo;
    }
}
```

Then `ActionsObservable` is free to define overrides that specifically apply when `T extends Action` constraint, which is what it should be all about, right?

pelotom:
> I created a similar PR (#288) which does essentially the same thing, but your PR seems to take into account a bunch of wider considerations than mine, so I'd say go with yours @pelotom

Hey @philcockfield, I think the main advantage of this approach is that types involved can be completely inferred without relying on a (possibly faulty, or missing) type annotation.

index.d.ts
jayphelps: I don't think we should add them to the Observable prototype because that leaks redux-observable stuff into all rxjs which in practice _probably_ won't be a big deal but is still not a great thing for a library like redux-observable to do (unless of course that's the [only point of the library](https://github.com/cartant/rxjs-etc)). People also use things we do as proof it's the "right way" so we gotta be careful.

pelotom: I guess I'm having a failure of imagination... what's an example of a `redux-observable`-only method which _doesn't_ assume that the value of type of the observable is an action, but which you wouldn't want to be be generally available on all `Observable`s in the app?

pelotom: Yeah, that makes sense and I agree. So given that constraint, I'm pretty happy with having `ActionsObservable<R extends Action>`, and then when you need to override an method which returns an `Observable`, use the double-overload pattern I did with `lift`:

```ts
lift<R extends Action>(operator: Operator<T, R>): ActionsObservable<R>;
lift<R>(operator: Operator<T, R>): Observable<R>;
```

That means that as long as your value type continues to be an action, the type system will still be aware that you're dealing with `ActionsObservables`--otherwise, you'll have to provide a cast. This seems like it will work with no extra effort in the vast majority of cases, and it gives a whole lot more type safety for things like `ofType` which only make sense if you're dealing with actions. What do you think?

jayphelps: I don't want people to need explicit casts. Is there a problem with my proposal?

pelotom: If you mean this proposal:

> This should still return and ActionsObservable (the same signature as it was before) or am I confused?

the yes, the problem is that it's impossible to keep the exact same signature as before if we have `ActionsObservable<T extends Action>`.

jayphelps: Hmm maybe you missed my proposal since I added it as an edit: https://github.com/redux-observable/redux-observable/pull/289#discussion_r132610733

pelotom: I'm not sure what you're proposing there?

jayphelps: Don't worry about it. I think we can cross that bridge when/if we ever need to.

jayphelps:
Gonna give this a day or two, if no further discussion from community, mergin! 🌮

jayphelps:
@pelotom this will require a rebase when you get a chance--make sure not to blow away the recent changes from #250

pelotom:
This should be ready to go now btw

jayphelps:
Yay! 🎉 🌮 🌮 🌮  Thanks again for all your work on this!! I'm going to cut a release tomorrow assuming no one reports any issues.

jayphelps:
Released as 0.16.0 🎈

pelotom:
Thanks a bunch!

PR #295, opened 2017-08-07
docs(examples): porting of the redux saga shopping cart example

jesinity:

### DIFF
--- a/examples/redux-observable-shopping-cart/js/epics/index.js
+++ b/examples/redux-observable-shopping-cart/js/epics/index.js
@@ -1,0 +1,49 @@
+import 'rxjs';
+import { Observable } from 'rxjs/Observable';
+import { combineEpics } from 'redux-observable';
+import { getCart } from '../reducers';
+import { api } from '../services';
+import * as actions from '../actions';
+
+/**
+ * search products epic
+ * @param action$
+ * @param store
+ * @returns {any|*|Observable}
+ */
+const searchProducts = (action$) => {
+  return action$
+    .ofType(actions.GET_ALL_PRODUCTS).switchMap(q => {
+      /*
+       * This example uses the same api of the redux-saga example, thus
+        * it adapts a Promise to an Observable.
+       */
+      return api.getProducts().map(actions.receiveProducts)
+    });
+};
+
+/**
+ * checkout epic.
+ * @param action$
+ * @param store
+ * @returns {any|*|Observable}
+ */
+const checkout = (action$,store) => {
+  return action$
+    .ofType(actions.CHECKOUT_REQUEST).switchMap(q => {
+      const cart = getCart(store.getState());
+      return api.buyProducts(cart)
+        .map(cart => actions.checkoutSuccess(cart))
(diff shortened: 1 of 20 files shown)


examples/redux-observable-shopping-cart/js/services/index.js
```diff
@@ -0,0 +1,26 @@
...
+const TIMEOUT = 100;
+const MAX_CHECKOUT = 2; // max different items
+
+export const api = {
+  getProducts() {
+    return new Promise( resolve => {
+      setTimeout(() => resolve(_products), TIMEOUT)
+    })
+  },
+
```
jayphelps: I'm thinking we should use Observables instead of Promises everywhere to set a good example for consistency/easy/cancelability.

e.g.

```js
getProducts() {
  return new Observable(observer => {
    const timerId = setTimeout(() => {
      observer.next(_products);
      observer.complete();
    }, TIMEOUT);

    return () => clearTimeout(timerId);
  });
}
```

What do you think?

examples/redux-observable-shopping-cart/js/epics/index.js
```diff
@@ -0,0 +1,52 @@
+import 'rxjs';
+import { Observable } from 'rxjs/Observable';
+import { ajax } from 'rxjs/observable/dom/ajax';
```
jayphelps: this is unused AFAICT

examples/redux-observable-shopping-cart/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "name": "redux-saga-shopping-cart",
+  "version": "1.0.0",
+  "main": "index.js",
+  "scripts": {
+    "webpack": "webpack",
+    "start": "http-server"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
jayphelps: MIT license plz

examples/redux-observable-shopping-cart/js/services/index.js
jesinity: It was intentional to leave it that way because the original example uses promises, so I tried to change only the middleware leaving the other code intact.
Maybe I could also add the original code so that a user could swap the epic with the saga and taste the difference.
The Observable since the beginning really rocks though... :-)  tell me what you think it's best I can change the code if you prefer.

jayphelps: Hmm yeah I still think it should be an Observable since we're not just changing middleware--we use rxjs operators and rxjs is not redux-observable. It's best to provide examples as idiomatic rxjs as possible, even if that means it differs from redux-saga in implementation.

jesinity: +1 ok, will do it.

jayphelps:
Thank you very much @jesinity for taking the time! I hope to see you around again. 😄

jesinity:
yeah! sorry I would have pushed again today the PR, on sunday I run out of time. 
BTW I pushed a new one from a new branch so, if it is useful, at least I will follow the correct workflow.

PR #302, opened 2017-08-09
docs(examples): porting of the redux saga cancellable counter example

jesinity:
I've ported the "redux saga cancellable counter example".
The porting is not 1:1 in terms of middleware adaptations, as I needed to use an extra action. I kept the original saga and the redux saga dependencies as well, so that an user could easily swap the 2 flavours of the examples if he wants to in the `webpack.config.js`. 
If you think this could be misleading I could remove the redux saga part.

Thank you for this fantastic project!

### DIFF
--- a/examples/redux-observable-cancellable-counter/.babelrc
+++ b/examples/redux-observable-cancellable-counter/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["es2015", "stage-0", "react"]
+}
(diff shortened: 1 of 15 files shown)


jayphelps:
Thanks for your work! I will probably indeed remove the saga stuff when I get a chance to think about it more.

jesinity:
Thank you!
I think as well the saga stuff may help only from a porting perspective, and removing it could provide a more elegant solution.
What could be improved I think is that the saga end the epic do not match 1 to 1: from this perspective, if you want to keep them together, it would be great that they would be triggered by the same actions.
Then I think the counter stuff could be improved as well, I don't feel it's super elegant.
Anyway in examples like this, where the flow is already non trivial, you see that redux-observable management of actions really shine through compared to redux-saga, that I find a bit complicated to follow.

PR #327, opened 2017-09-15
docs(examples): Remove deprecated React.PropTypes in favor of the pro…

iamclaytonray:
`fix(PropTypes): Remove deprecated React.PropTypes in favor of the prop-types pkg`

<a href='https://facebook.github.io/react/docs/typechecking-with-proptypes.html'>Facebook reference here</a>

This PR just removes PropTypes from the React lib and moves it into `prop-types` as recommended by Facebook, and removes the `console.warning()` notice about the deprecation. I also added a couple of missing dependences: `rxjs` and `babel-preset-stage-0`. ***This is for `examples/` only***

### DIFF
--- a/examples/redux-observable-cancellable-counter/js/components/Counter.js
+++ b/examples/redux-observable-cancellable-counter/js/components/Counter.js
@@ -1,5 +1,6 @@
 /*eslint-disable no-unused-vars*/
-import React, { PropTypes } from 'react'
+import React from 'react'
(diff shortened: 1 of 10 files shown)


iamclaytonray:
Second commit was to actually add the prop-types package. Some days, I just don't get enough coffee... 😞

iamclaytonray:
Cool, I'll keep that in mind for the future

PR #329, opened 2017-09-23
chore: rename (es -> esm)

Shyam-Chen:
ref, https://github.com/vuejs/vue/blob/dev/package.json#L6


jayphelps:
yeah esm is more clear. Thanks! 🎉

Issue #263, opened 2017-06-21
Reducer errors swallowed silently after epics' outputs


haf:
```
const reducer = (state = {}, action) => {
  switch (action.type) {
    case 'THING':
      return { ...state, thing: action.nonexistent.childprop }
    default:
      return state;
  }
};
export function sampleEpic(action$: Observable, store: Store) {
  return action$.ofType(ANYTHING)
    .concatMap(action =>
      fetch('/api/anything', { ... })
    .mergeMap(response => {
      switch (response.status) {
        case 409:
          return Observable.of(pleaseNo(email));
        case 201:
          return Observable.of({ type: 'THING' }, { type: 'NEVER_DISPATCHED' });
        default:
          throw new Error(`Unknown response code ${response.code}`);
      }
    });
}
```

The THING action causes a crash that is hidden from the user (me for the last 2 hours) so that it's very hard to see where things went wrong.

Adding `catch` to the end of the epic doesn't make a difference.

All middleware a successfully run.

BerkeleyTrue:
I can't see how redux-observable would catch errors thrown in your reducer. RO does not contain your reducer or any other redux aspect other than the epics themselves. Also, there aren't cases where RxJS would swallow errors like promises would. The default is always to throw errors if unhandled by the user.

I'm also failing to see how the `THING` action would crash as there are no syntax errors present
```js
 return { ...state, thing: action.nonexistent }
```
The result of this line of code is a state with the `thing` key set to `undefined`.

haf:
Perhaps it's a redux bug then?

`.childprop` in sample now.

jayphelps:
I'm digging into this. Right now my findings are that RxJS did not swallow this error in 5.0.3 but started to in 5.1.0. https://jsbin.com/gehasuy/edit?html,js,output

It's not yet clear if this is a bug in Rx or if we were relying on buggy behavior that was fixed in 5.1.0. My initial (biased) reaction is that Rx should not be swallowing this error condition but it's possible I'm missing something.

Will continue digging.

jayphelps:
Minimum reproduction:

```js
const input$ = new Rx.Subject();

input$
  .mergeMap(() => Observable.of(1, 2))
  .subscribe(d => {
    throw new Error('some error');
  });

input$.next();
```

Also appears related to Subject usage (but could be a red herring)

jayphelps:
[...]

PR #332, opened 2017-09-28
docs(troubleshooting): Side effects with subscribe

### LINKED ISSUE #331: Troubleshooting docs: epics fail silently when returning subscribe from epic
When you return a subscribe from an epic, it just fails silently and none of the epics execute. I'm going to submit a PR with an example on how to do side effects without using `subscribe`.

hendrikswan:
closes #331

### DIFF
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -109,11 +109,10 @@ const myEpic = action$ => {
 
 The problem with this is that calling `subscribe` on an `Observable` does not return another `Observable`.
 All epics should return an `Observable`.
+If you ran into this issue and want to learn more about it, read through this issue: https://github.com/redux-observable/redux-observable/issues/263.
 
 If you want your epic to be "read-only", meaning you want it to perform side-effects
 without producing any downstream actions, you can use the following pattern.
-
-This approach returns an empty `Observable` from the epic.
 
 ```js
 const myEpic = action$ => {
@@ -124,7 +123,7 @@ const myEpic = action$ => {
 };
 ```
 
-
+This approach esdsentially returns an empty `Observable` from the epic, which does not cause any downstream actions.
 
 
 * * *
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -95,6 +95,28 @@ class TooFancy {
 
 See https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Arrow_functions (Arrow functions used as methods)
 
+### It seems like the epics are not executing 
+
+You might be doing a `subscribe` in one of your epics: 
+
+```js
+const myEpic = action$ => { // performing side effect with .subscribe!
+  return action$.ofType(...)
+  .subscribe(item => console.log(item));
+};
+```
+
+This happens because subscribe doesn't return a promise itself. When you want to perform side effects based on actions, without the need to return an observable from your epic containing downstream actions, you can follow this pattern:
+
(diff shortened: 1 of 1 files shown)


docs/Troubleshooting.md
```diff
@@ -95,6 +95,28 @@ class TooFancy {
...
+You might be doing a `subscribe` in one of your epics: 
+
+```js
+const myEpic = action$ => { // performing side effect with .subscribe!
+  return action$.ofType(...)
+  .subscribe(item => console.log(item));
+};
+```
+
+This happens because subscribe doesn't return a promise itself. When you want to perform side effects based on actions, without the need to return an observable from your epic containing downstream actions, you can follow this pattern:
```
jayphelps: > This happens because subscribe doesn't return a promise itself.

Can you clarify what you mean here?

> without the need to return an observable from your epic containing downstream actions

Epics must always return an Observable, and indeed you are--it's just an Observable that never emits anything. An important distinction.

jayphelps:
Cross posting here: https://github.com/redux-observable/redux-observable/issues/331#issuecomment-332957470

> This is unfortunately another example of #263 and https://github.com/ReactiveX/rxjs/issues/2813. There is an error about not providing an Observable where it was expected, it's just being swallowed silently due to a bug.
> 
> As https://github.com/ReactiveX/rxjs/issues/2813 mentions though, while this is a bug, in the future this behavior will likely be intentional, though almost certainly with a better debugging experience by still showing the error in the debug console (which is what most people actually care about when they freak out about this change lol).

hendrikswan:
Yeah, good idea, I'll link to the reason why it happens.

hendrikswan:
Hey @jayphelps, I linked to the issue as you suggested. Everything you linked to seems to expound on swallowing of errors. 

Maybe I'm missing something, and I wonder if it also wouldn't overwhelm other `Observable` laymen like me, but why didn't my scenario just result in this error? `TypeError: object is not observable`. 

It's as if the result of calling `subscribe` looks enough like an `Observable` and fools the logic that checks whether it is one, and we then proceed to work with it as if it's an `Observable`, which in turn causes errors to be swallowed. 

Isn't there something that we can do in redux-observable, so that we can catch this scenario and provide a specific error? Something like: "You did not return an observable from your epic".

hendrikswan:
And if there is something that we can do, I'd be happy to do a PR for that.

jayphelps:
Thanks @hendrikswan ! 🎉

hendrikswan:
Thanks for your patience @jayphelps :)

jayphelps:
> Maybe I'm missing something, and I wonder if all this information wouldn't also overwhelm other Observable laymen like me, but why didn't my scenario just result in this error? TypeError: object is not observable.

@hendrikswan oh shoot I wrote a nice reply to that question... I must have forgot to send it.. Sorry about that.

tl;dr is that there is in fact an error, it's just being [silently swallowed by rxjs](https://github.com/ReactiveX/rxjs/issues/2813). You can see the error by placing a `.do({ error: e => console.error(e) })` in your root epic

```js
const rootEpic = (...args) =>
  combineEpics(...yourEpicsHere)(...args)
    .do({ error: e => console.error(e) });
```

The error is just the normal:

> You provided and invalid object where a stream was expected. You can provide an Observable, Promise, Array, or Iterable.

hendrikswan:
@jayphelps, ah now I understand! Thanks a lot for taking the time to explain it 👍

PR #336, opened 2017-10-10
feat(EpicMiddleware): warn about improper use of dispatch function

### LINKED ISSUE #314: RFC: Removing the store.dispatch provided to epics
Hey all, when we originally made redux-observable we thought it would be a good idea to keep the ability to imperatively call `store.dispatch` inside your epics as basically a rarely used (if ever) escape hatch to just "get things done". While it would be nice to have it _just in case_, I'm quite concerned with the number of new people I come across who use it exclusively, rather than emitting actions. e.g. I see people calling it inside `map` super often and I feel like it's our fault for giving people the gun to shoot themselves with. I see this in stackoverflow, gitter, and even youtube tutorials on _how to use redux-observable_. 😢

Sooo. I'm proposing we remove it entirely from epics. I'm not sure what that means for the function signature of epics but I imagine it would change since the second argument object would only have getState now and that's just weird. If we generally have consensus to remove dispatch, having it as part of the [store enhancer effort](https://github.com/redux-observable/redux-observable/pull/259) since that will provide a cleaner migration/deprecation path for people. If that's the case, both of these and a few other minor things would probably be part of us finally starting to move towards a 1.0. Starting with 1.0.0-alpha release first, then beta, etc.

Thoughts? I'm definitely willing to hear out people's arguments for keeping it.

Cc/ @BerkeleyTrue @rgbkrk

evertbouw:
closes #314

### DIFF
--- a/src/createEpicMiddleware.js
+++ b/src/createEpicMiddleware.js
@@ -31,12 +31,20 @@ export function createEpicMiddleware(epic, options = defaultOptions) {
   const epicMiddleware = _store => {
     store = _store;
 
+    const vault = (process.env.NODE_ENV === 'production') ? store : {
+      getState: store.getState,
+      dispatch: (action) => {
+        store.dispatch(action);
+        console.warn(`Your Epic "${epic.name || '<anonymous>'}" called store.dispatch directly. This is an anti-pattern.`);
+      }
+    };
+
     return next => {
       epic$
         ::map(epic => {
           const output$ = ('dependencies' in options)
-            ? epic(action$, store, options.dependencies)
-            : epic(action$, store);
+            ? epic(action$, vault, options.dependencies)
+            : epic(action$, vault);
 
           if (!output$) {
             throw new TypeError(`Your root Epic "${epic.name || '<anonymous>'}" does not return a stream. Double check you\'re not missing a return statement!`);
--- a/src/createEpicMiddleware.js
+++ b/src/createEpicMiddleware.js
@@ -31,17 +31,17 @@ export function createEpicMiddleware(epic, options = defaultOptions) {
   const epicMiddleware = _store => {
     store = _store;
 
-    const vault = (process.env.NODE_ENV === 'production') ? store : {
-      getState: store.getState,
-      dispatch: (action) => {
-        store.dispatch(action);
-        console.warn(`Your Epic "${epic.name || '<anonymous>'}" called store.dispatch directly. This is an anti-pattern.`);
-      }
-    };
-
     return next => {
       epic$
         ::map(epic => {
+          const vault = (process.env.NODE_ENV === 'production') ? store : {
+            getState: store.getState,
+            dispatch: (action) => {
+              store.dispatch(action);
+              console.warn(`Your Epic "${epic.name || '<anonymous>'}" called store.dispatch directly. This is an anti-pattern.`);
+            }
+          };
+
           const output$ = ('dependencies' in options)
             ? epic(action$, vault, options.dependencies)
             : epic(action$, vault);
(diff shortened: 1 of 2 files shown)


src/createEpicMiddleware.js
```diff
@@ -31,12 +31,20 @@ export function createEpicMiddleware(epic, options = defaultOptions) {
   const epicMiddleware = _store => {
     store = _store;
 
+    const vault = (process.env.NODE_ENV === 'production') ? store : {
+      getState: store.getState,
+      dispatch: (action) => {
+        store.dispatch(action);
+        console.warn(`Your Epic "${epic.name || '<anonymous>'}" called store.dispatch directly. This is an anti-pattern.`);
```
jayphelps: `epic.name` refers to the root epic in this scope I think? `export function createEpicMiddleware(epic, ...` If so, it'll probably be confusing. You probably had this code originally inside the `::map(epic => {` scope, in which case `epic` refers to the actual epic that did use `store.dispatch`, but you'd have to create a "vault" for each epic, which I think is an ok tradeoff. 😄 

While you're in here, you might change the root epic variable from `epic` to `rootEpic` to make it less confusing and prevent our `::map(epic => {` from shadowing it.

test/createEpicMiddleware-spec.js
```diff
@@ -20,13 +22,29 @@ describe('createEpicMiddleware', () => {
...
-      expect(epic.firstCall.args[1]).to.equal(store);
+      expect(epic.firstCall.args[1].getState).to.equal(store.getState);
       done();
     };
     const store = createStore(reducer, applyMiddleware(epicMiddleware, mockMiddleware));
     store.dispatch({ type: 'FIRST_ACTION_TO_TRIGGER_MIDDLEWARE' });
   });
 
+  it('should warn about improper use of dispatch function', () => {
+    sinon.spy(console, 'warn');
```
jayphelps: Missing a matching restore of the original at the end of the test, `console.warn.restore()`

test/createEpicMiddleware-spec.js
```diff
@@ -20,13 +22,29 @@ describe('createEpicMiddleware', () => {
...
+      .ofType('PING')
+      ::map(() => store.dispatch({ type: 'PONG' }))
+      ::ignoreElements();
+
+    const middleware = createEpicMiddleware(epic);
+    const store = createStore(reducer, applyMiddleware(middleware));
+
+    store.dispatch({ type: 'PING' });
+
+    expect(console.warn.called).to.equal(true);
```
jayphelps: Can you change to `expect(console.warn.callCount).to.equal(1);` or something similar--just want to make sure console.warn is only called once.

src/createEpicMiddleware.js
evertbouw: Ah yes. Maybe I can warn the action.type instead to make the message more helpful?

evertbouw: Hmm I moved the vault into the .map but it seems the epic there is the root epic, and the name undefined. Maybe I can use the action.type in the warning message instead?

(I chose "vault" as a name for a store with more restrictions)

jayphelps:
🎉  woot! thank you @evertbouw!! Very much appreciated.

src/createEpicMiddleware.js
jayphelps: Ohhh right, I forgot that inside the middleware we don't have access to the individual epics because they were composed with `combineEpics`. I think I'll adjust it to just be generic and not mention which specific epic is to blame. I'm also going to change the message a bit to clarify that we're actually deprecating this (store.dispatch inside epics) and it will be removed in v1.0.0.

PR #186, opened 2017-02-03
Moving ofType into a separate operator

jshthornton:
#185 

- [X] I have read the [Contributor Guide](https://github.com/redux-observable/redux-observable/blob/master/CONTRIBUTING.md)
- [X] My commit messages are in [conventional-changelog-standard](https://github.com/redux-observable/redux-observable/blob/master/CONTRIBUTING.md#sending-a-pull-request) format.

PR #343, opened 2017-10-25
feat(ofType): expose ofType as lettable operator

jgoz:
Supports using `ofType` directly in lettable operator pipelines.

Closes #186


jgoz:
Happy to contribute! Thanks for the quick turnaround 😃

rgbkrk:
Thank you so much!

BerkeleyTrue:
@jayphelps when will this be released?

jayphelps:
@BerkeleyTrue there ended up being an issue with it with versions of rxjs pre-lettable. I'm still debugging.

jayphelps:
I think the issue is actually not version, but that the global version of rxjs doesn't provide let as `letProto`, it provides it as `letBind`. I dunno, shit is confusing.

jayphelps:
That is indeed the issue. The named export is called `letProto` but the prototype name is `letBind`, so webpack-rxjs-externals doesn't correctly resolve it for UMD. Tracking here: https://github.com/ReactiveX/rxjs/issues/3024

Not sure of how to work around..even if a new rxjs version fixes this, we wouldn't be able to support older versions of rxjs which is a deal breaker IMO. Anyone have ideas?

jgoz:
@jayphelps Hmm, I wasn't aware of `webpack-rxjs-externals`. We can't go back in time and rename `letProto` to `letBind`, so maybe the simplest solution is to copy the implementation back into `ActionsObservable.ofType` instead of using the operator directly.

Other options:
- Inside `ActionsObservable.js` do `import 'rxjs/add/operator/let'` and use `let`/`letBind` directly (downside being it pollutes `Observable` for users, but `let` is super tiny...)
- Expose `ofType` as a bindable operator too (`ofTypeProto`? using the lettable version internally) and binding to that in `ActionsObservable`

jayphelps:
lol what I am thinking...let is so simple we don't even need to use it. 😄 

```js
return ofType(...keys)(this)
```

I'll make this change.

jgoz:
Haha, that is by far the best option and I wish I'd thought of it initially 😉

jayphelps:
#345 fixes that issue. I also want to get #346 reviewed and merged before we release the next version. Reviews from anyone are welcome!

jayphelps:
Released in v0.17.0 🎈

PR #345, opened 2017-10-30
fix(ofType): don't depend on letProto as it has issues in UMD builds

jayphelps:
Cc/ @BerkeleyTrue @jgoz


jayphelps:
Thanks for the reviews all! Going to release this right now.

PR #346, opened 2017-10-30
fix(store.dispatch): Calling `store.dispatch()` directly in your Epic is now deprecated

jayphelps:
The ability to call `store.dispatch()` inside your Epics was originally provided as an escape hatch, to be used rarely, if ever. Unfortunately in practice we've seen a large number of people using it extensively; there has even been popular tutorials teaching it as how you use redux-observable. Instead, Epics should emit actions through the Observable the Epic returns, using idiomatic RxJS.

```js
const somethingEpic = (action$, store) =>
  action$.ofType(SOMETHING)
    .switchMap(() =>
      ajax('/something')
        .do(() => store.dispatch({ type: SOMETHING_ELSE }))
        .map(response => ({ type: SUCCESS, response }))
    );
```

```js
const somethingEpic = action$ =>
  action$.ofType(SOMETHING)
    .switchMap(() =>
      ajax('/something')
        .mergeMap(response => Observable.of(
          { type: SOMETHING_ELSE },
          { type: SUCCESS, response }
        ))
    );
```

`store.dispatch` will be removed from Epics in v1.0.0 of redux-observable. This is unrelated to usage of `store.dispatch` inside your UI components--you will continue to use it there

Cc/ @evertbouw

### DIFF
--- a/docs/basics/Epics.md
+++ b/docs/basics/Epics.md
@@ -202,7 +202,7 @@ dispatch(incrementIfOdd());
 ```
 > REMEMBER: When an Epic receives an action, it has already been run through your reducers and the state updated.
 
-Using `store.dispatch()` inside your Epic is a handy escape hatch for quick hacks, but use it sparingly. It's considered an anti-pattern and we may remove it from future releases.
+Using `store.dispatch()` inside your Epic is an anti-pattern and will be removed in v1.0.0 of redux-observable so it's best not to use it.
 
 ***
 
--- a/src/createEpicMiddleware.js
+++ b/src/createEpicMiddleware.js
@@ -37,7 +37,7 @@ export function createEpicMiddleware(rootEpic, options = defaultOptions) {
           const vault = (process.env.NODE_ENV === 'production') ? store : {
             getState: store.getState,
             dispatch: (action) => {
-              console.warn(`Your Epic "${epic.name || '<anonymous>'}" called store.dispatch directly. This is an anti-pattern.`);
+              require('./utils/console').deprecate('calling store.dispatch() directly in your Epics is deprecated and will be removed. Instead, emit actions through the Observable your Epic returns.\n\n  https://goo.gl/WWNYSP');
               return store.dispatch(action);
             }
           };
--- a/src/utils/console.js
+++ b/src/utils/console.js
@@ -1,0 +1,10 @@
+const deprecationsSeen = {};
+
+export const deprecate = (typeof console === 'object' && typeof console.warn === 'function')
+  ? msg => {
+    if (!deprecationsSeen[msg]) {
+      deprecationsSeen[msg] = true;
+      console.warn(`redux-observable | DEPRECATION: ${msg}`);
+    }
+  }
+  : () => {};
--- a/test/createEpicMiddleware-spec.js
+++ b/test/createEpicMiddleware-spec.js
@@ -43,6 +43,7 @@ describe('createEpicMiddleware', () => {
     store.dispatch({ type: 'PING' });
 
     expect(console.warn.callCount).to.equal(1);
+    expect(console.warn.getCall(0).args[0]).to.equal('redux-observable | DEPRECATION: calling store.dispatch() directly in your Epics is deprecated and will be removed. Instead, emit actions through the Observable your Epic returns.\n\n  https://goo.gl/WWNYSP');
     console.warn.restore();
   });
 


src/createEpicMiddleware.js
```diff
@@ -37,7 +37,7 @@ export function createEpicMiddleware(rootEpic, options = defaultOptions) {
           const vault = (process.env.NODE_ENV === 'production') ? store : {
             getState: store.getState,
             dispatch: (action) => {
-              console.warn(`Your Epic "${epic.name || '<anonymous>'}" called store.dispatch directly. This is an anti-pattern.`);
+              require('./utils/console').deprecate('calling store.dispatch() directly in your Epics is deprecated and will be removed. Instead, emit actions through the Observable your Epic returns.\n\n  https://goo.gl/WWNYSP');
```
jayphelps: This is so that when `process.env.NODE_ENV === 'production'` we don't import the console.js file at all since it's not needed.

rgbkrk: I like this pattern.

src/utils/console.js
```diff
@@ -0,0 +1,10 @@
+const deprecationsSeen = {};
+
+export const deprecate = (typeof console === 'object' && typeof console.warn === 'function')
+  ? msg => {
+    if (!deprecationsSeen[msg]) {
+      deprecationsSeen[msg] = true;
+      console.warn(`redux-observable | DEPRECATION: ${msg}`);
+    }
```
rgbkrk: Thanks for making it back off after being logged at least once.

docs/basics/Epics.md
```diff
@@ -202,7 +202,7 @@ dispatch(incrementIfOdd());
 ```
 > REMEMBER: When an Epic receives an action, it has already been run through your reducers and the state updated.
 
-Using `store.dispatch()` inside your Epic is a handy escape hatch for quick hacks, but use it sparingly. It's considered an anti-pattern and we may remove it from future releases.
+Using `store.dispatch()` inside your Epic is an anti-pattern and will be removed in v1.0.0 of redux-observable so it's best not to use it.
```
rgbkrk: I like how this goes from `handy escape hatch` to `anti-pattern`.

jayphelps:
Don't merge yet, I forgot to fix the tests! DOH

jayphelps:
Ok once the build passes we're good to go.

jayphelps:
Thanks for the review @rgbkrk!!

adred:
I'm new to rxjs. So does this mean I won't be able to call action creators anymore? It sucks because instead of reusing action creators I now have to type the actions. :-(

jayphelps:
@adred you can ABSOLUTELY use action creators. This only deprecated using the store.dispatch function inside epics, that’s it.

adred:
@jayphelps yeah, that's what I meant. I'm using `dispatch(actionCreator)` inside my epics.

jayphelps:
@adred gotcha, different terminology.

It’s very anti redux-observable and was never intended to be used except for hacks—which a person can still do, they can inject the store themselves, or export and import it, etc. I recommend against it though, as this is fighting the redux-observable way. There’s no harm in not liking it, there are lots of middleware to choose from or you can roll your own. 🖖

rgbkrk:
If you need to dispatch in the middle of your epic, there are many ways to do it. By the nature of the epic you need to emit actions anyways. You can merge together multiple streams in tandem.

```js
import { of } from "rxjs/observable/of";
import { empty } from "rxjs/observable/empty";
import { mergeMap } from "rxjs/operators";

action$.ofType("SOME_TYPE").pipe(
  mergeMap(action => {
    // within merge map you return an observable of actions
    // and they become the next part of the stream
    if(...) {
      // You can return one
      return of(someAction);
    } else if {
      // or many
      return of(someAction, anotherAction);
    } else {
      // or nothing
      return empty();
    }
   // or even whole new streams composed of various things
  })
)
```

trubi:
Guys, quick question.
Sometimes I am using this (with native module):
```
const magtekModule = NativeModules.MagtekManager
const eventEmitter = new NativeEventEmitter(magtekModule)

eventEmitter.addListener(
    "MagtekManagerEventDataReady",
    (payload) => dispatch({
            type: MAGTEK_MANAGER_DATA_READY,
            payload: {
                magstripe: payload[0],
        }
    }))
```

Any alternative? Thank!

jayphelps:
@trubi happy to help if you wanna post it on StackOverflow so the knowledge is shared.

trubi:
Great idea indeed thank! https://stackoverflow.com/questions/47665255/alternative-to-direct-calling-of-store-dispatch

ecnaidar:
So the getState() will still be available as a parameter, right? Removing dispatch is a sane move to make people do thing Rx way, but removing ability to access state might make some things unnecessarily complicated.
I will leave what I have came up with here just in case. No idea if that is a 100% correct way of dispatching in the middle of epic. (Not really experienced with RxJS but worked for a while with redux-saga)

``` javascript
action$
    .ofType(CREATE)
    .flatMap(action => {
      // Save our request observable so we can process output two different ways
      const request$ = Observable.fromPromise(requestCreate(action.payload));
      return [
        // Before API request 
        Observable.of(startSubmit("template")),
        // 1. Notify that load is done
        request$.map(response => {
          console.log("stopping submit");
          return stopSubmit("template", response.error || {});
        }),
        // 2. Format and emit result
        request$.map(response => {
          console.log("returning response");
          return saveTemplate(response);
        })
      ];
      // Each element produces plain redux action
    })
    .concatAll();
```

Issue #192, opened 2017-02-11
[Feature request] Support for stringifyable action factories


thezanke:
I think it would be immensely helpful if `action$.ofType()` knew how to handle stringifyable action factories like those that you see from redux-actions or redux-act. This is helpful for people who, following the guidance of the React docs, have picked up action creator factory libraries to reduce boilerplate. One of the major pieces of boilerplate these packages eliminate is the need to maintain a list of constants because all action factories have a `.toString()` method that return their type so they can be used as the action type as well as the action creator.

This would be easy to handle, especially since there is no reason for calling `ofType()` on a function as it is. The quickest way to implement this in this case would probably be just calling `String()` on `keys[i]` when comparing it against `type`.

This would allow for something like the following:
```
import { action1, action2 } from 'actions/bots';

export default (action$, store) =>
  action$.ofType(action1)
    .do(//something)
    .map(() => action2());
```

I'll whip up a quick PR.

thezanke:
I just noticed that this has been discussed before and I just didn't see it when I looked (used wrong terms). The fear is breaking support for using symbols as a type. Could we just have a simple matcher that just tests if `typeof key[i] === 'function'` before calling `String()` or `.toString()` on it? This would really help clean up the need to use constants file and makes it so I don't have to wrap with `String()` myself every single time I use `ofType()`.

jayphelps:
Hey @theengineear! Thanks for the ticket + PR--bringing code to the table is always super appreciated!!

> I just noticed that this has been discussed before

Yeah, just for posterity, it has previously been proposed in #71, then PR'd in #175

> Could we just have a simple matcher that just tests if typeof key[i] === 'function' before calling String() or .toString() on it? 

This idea seems promising. As long a test was added that showed Symbols were supported still too I think this is solid

PR #348, opened 2017-10-31
feat(ofType): integration with redux-actions.

theengineear:
The actionCreators you make with redux-actions implement their own
toString, which means you don't have to keep track of action types at
all. Before this commit, you need to:

```
ofType(myActionCreator.toString())
```

In order to get this to work as expected. As of this commit, you can:

```
ofType(myActionCreator)
```

Closes: #192

### DIFF
--- a/test/operators-spec.js
+++ b/test/operators-spec.js
@@ -52,5 +52,71 @@ describe('operators', () => {
       expect(lulz).to.deep.equal([{ type: 'LULZ', i: 0 }, { type: 'LARF', i: 1 }]);
       expect(haha).to.deep.equal([{ type: 'HAHA', i: 0 }]);
     });
+
+    it('should handle actionCreators which define a toString method', () => {
+      // This helps when using the popular `redux-actions` npm package.
+      const actions = new Subject();
+      const cache1 = [];
+      const cache2 = [];
+      const LULZ_TYPE = 'LULZ';
+      const HAHA_TYPE = 'HAHA';
+      const LARF_TYPE = Symbol();
+
+      const createActionCreator = type => {
+        const actionCreator = payload => ({ type, payload });
+        actionCreator.toString = () => type;
+        return actionCreator;
+      };
+
+      const lulz = createActionCreator(LULZ_TYPE);
+      const haha = createActionCreator(HAHA_TYPE);
+      const larf = createActionCreator(LARF_TYPE);
+
+      // Sanity check.
+      expect(String(lulz)).to.deep.equal(LULZ_TYPE);
+      expect(lulz(0)).to.deep.equal({ type: LULZ_TYPE, payload: 0 });
+
+      actions.pipe(ofType(lulz, larf)).subscribe(x => cache1.push(x));
+      actions.pipe(ofType(haha)).subscribe(x => cache2.push(x));
+
+      actions.next(lulz(0));
+      expect(cache1).to.deep.equal([lulz(0)]);
+      expect(cache2).to.deep.equal([]);
+
+      actions.next(larf(1));
+      expect(cache1).to.deep.equal([lulz(0), larf(1)]);
+      expect(cache2).to.deep.equal([]);
+
+      actions.next(haha(0));
+      expect(cache1).to.deep.equal([lulz(0), larf(1)]);
+      expect(cache2).to.deep.equal([haha(0)]);
+    });
+
+    it('should not fail when types are symbols', () => {
+      const actions = new Subject();
+      const cache1 = [];
+      const cache2 = [];
+      const LULZ_TYPE = Symbol();
+      const HAHA_TYPE = Symbol();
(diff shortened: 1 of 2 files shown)


theengineear:
I'd also really appreciate support for toString implementations. If someone wants to detail the concern surrounding `Symbol`s mentioned in #192, I'd be happy to incorporate that into the spec.

jayphelps:
Hey @theengineear thanks for the PR! I think if we only applied toString when the type is a function, this will continue to work for Symbols too.

Something like..

```js
const key = keys[i];
if (key === type || (typeof key === 'function' && key.toString() === type)) {

}
```

theengineear:
@jayphelps yup 👌 , more or less what I'm doing:

```
const keyHasType = (type, key) => {
  return type === key || typeof key === 'function' && type === String(key);
};
```

jayphelps:
@theengineear hmm I'm not seeing that code. Did you forgot to push perhaps?

image

theengineear:
@jayphelps just locking down the spec. I realized that if the toString implementation of an actionCreator returns a Symbol, you will get an error converting a Symbol to a string. I'll ping you when it's up/done.

test/operators-spec.js
```diff
@@ -52,5 +52,82 @@ describe('operators', () => {
...
+      // Sanity check.
+      expect(HAHA_TYPE).not.to.equal(LULZ_TYPE);
+      expect(LULZ_TYPE).not.to.equal(LARF_TYPE);
+      expect(LARF_TYPE).not.to.equal(HAHA_TYPE);
+      expect(HAHA_TYPE.toString()).to.equal(LULZ_TYPE.toString());
+      expect(LULZ_TYPE.toString()).to.equal(LARF_TYPE.toString());
+      expect(LARF_TYPE.toString()).to.equal(HAHA_TYPE.toString());
+      expect(String(HAHA_TYPE)).to.equal(String(LULZ_TYPE));
+      expect(String(LULZ_TYPE)).to.equal(String(LARF_TYPE));
+      expect(String(LARF_TYPE)).to.equal(String(HAHA_TYPE));
```
theengineear: I was surprised to see that `String(Symbol())` doesn't throw an error. But...

```
const foo = () => {};
foo.toString = () => Symbol();
String(foo);
```

DOES throw an error. Is this something that I'm not understanding that I need to take into account?

theengineear:
@jayphelps OK. All set. However, it might be a bad idea to allow a function's `toString` to return a symbol for now. That's the last thing I could be hung up on. Thoughts?

test/operators-spec.js
jayphelps: `toString` should always return a string. If someone overrides it and does not return a string, that's their fault lol 😄

test/operators-spec.js
```diff
@@ -52,5 +52,82 @@ describe('operators', () => {
...
+    it('should not fail when types are symbols', () => {
+      const actions = new Subject();
+      const cache1 = [];
+      const cache2 = [];
+      const LULZ_TYPE = Symbol();
+      const HAHA_TYPE = Symbol();
+      const LARF_TYPE = Symbol();
+
+      // Sanity check.
+      expect(HAHA_TYPE).not.to.equal(LULZ_TYPE);
```
jayphelps: Curious..why are these sanity checks needed? I don't think we need to test Symbols themselves in contexts unrelated to `ofType`.

theengineear:
Some context. Looks like `redux-actions` has discussed dropping symbols support before https://github.com/reduxactions/redux-actions/issues/126

test/operators-spec.js
jayphelps: obv do let me know if you think they're needed 😄

theengineear: Ha, ya I can take them out. I thought it might be helpful to document why the test is there, but I think it's clear from the test string.

theengineear:
@jayphelps ok, updated :tada:

jayphelps:
So awesome. Thank you very much @theengineear!!!

jayphelps:
Released in v0.17.0 🎈

rgbkrk:
Thanks! I had some dangling comments in here, which I'll go ahead and post even though this is merged and I think we're good to go.

Even though `redux-actions` does it, `Symbol` so shouldn't be a return for `toString`:

* Symbols are unique `Symbol('a') !== Symbol('a')`, whereas `'a' === 'a'` for strings
* Coercing Symbols to strings leads to them no longer being Symbols whatsoever

```
> Symbol('a').toString() === Symbol('b').toString()
false
> Symbol('a').toString() === Symbol('a').toString()
true
```

Generally I'd say my statement on `toString` is that it should return a string, always. Not something that can be coerced (and Symbols won't cast to string without some force).

rgbkrk:
Thanks for this release @jayphelps, it's awesome.

PR #355, opened 2017-11-08
docs(ofType): clarify that `ofType` is variadic

jayphelps:
lotssssss of people come up with their own `.filter()` variant for this, even though it's always supported it but wasn't documented.


rgbkrk:
Thanks!

PR #361, opened 2017-11-13
feat(ofType): simplify code

BetterCallSky:

### DIFF
--- a/src/operators.js
+++ b/src/operators.js
@@ -1,23 +1,11 @@
 import { filter } from 'rxjs/operator/filter';
 
-const keyHasType = (type, key) => {
+const keyHasType = type => key => {
   return type === key || typeof key === 'function' && type === key.toString();
 };
 
 export function ofType(...keys) {
   return function ofTypeOperatorFunction(source) {
-    return source::filter(({ type }) => {
-      const len = keys.length;
-      if (len === 1) {
-        return keyHasType(type, keys[0]);
-      } else {
-        for (let i = 0; i < len; i++) {
-          if (keyHasType(type, keys[i])) {
-            return true;
-          }
-        }
-      }
-      return false;
-    });
+    return source::filter(({ type }) => keys.some(keyHasType(type)));
   };
 }


rgbkrk:
The commit message should be `refactor(ofType): simplify ofType by using Array.some`.

jayphelps:
The main reason I used a loop was to hopefully prevent people from complaining about the micro performance of `ofType`. I haven't tested this solution using `some`, but my intuition would guess it's significantly slower, relatively speaking. But in 99.9% of cases I can't imagine that micro perf difference mattering. If someone is dispatching enough actions for it to matter they'll prolly have bigger perf issues elsewhere.

Anyone have objections to this PR?

lsentkiewicz:
@rgbkrk 
fixed

@jayphelps 
Here is jsperf https://jsperf.com/some-vs-for-51546
loop: 220m ops/sec
some: 17m ops/sec

From my experience, if we dispatch more than 100 actions per sec it can cause performance issues in UI.

jayphelps:
> Here is jsperf https://jsperf.com/some-vs-for-51546
> loop: 220m ops/sec
> some: 17m ops/sec

Nice! Thank you for putting that together. Always better to have facts like these.

> From my experience, if we dispatch more than 100 actions per sec it can cause performance issues in UI.

I agree with the sentiment that emitting tons of actions continuously can often perform poorly, like if you used it for animations. I'm mostly referring to short-burst performance. e.g. I've seen thousands of actions dispatched in less than a second fairly often, each causing a UI update (even if not visually perceivable because the thread wasn't yielded to the browser to repaint between each).

He's a contrived example that chugs along at 1000 synchronous actions emitted "per second": https://jsbin.com/vubovey/edit?js,output Note the distinction here that the actions are emitted synchronously. So actually, a majority of the 1 second is spent doing nothing.

I'm not saying people _should_ necessarily think 1000s of actions is a good idea, just that at the end of the day we all have to shipit :shipit:  And in these cases, I just try to avoid giving people excuses to complain about how "redux-observable does xyz slow" 😄  Since most perf issues I see in the wild are "death by a thousand cuts" rather than a single bottleneck.

rgbkrk:
Given all this reasoning and the fact that nteract does pump pure messages through Rx + redux-observable, I'd prefer we keep the more performant `for` loop since it already works, is already tested, and has been confirmed to be faster, even if the code is a bit smaller here.

I'm completely surprised that `some` hasn't been completely optimized by Google in v8.

jayphelps:
@BetterCallSky Many years later but still thankful for your contribution! I think we're gonna stick with the tried and true for loop at the moment.

PR #380, opened 2017-12-05
fix(Reducer errors): Rethrow exceptions asynchronously instead of using console.error

jayphelps:
DO NOT MERGE. We're still discussing whether the existing solution or this are even viable. See #379

### DIFF
--- a/src/createEpicMiddleware.js
+++ b/src/createEpicMiddleware.js
@@ -10,7 +10,23 @@ const defaultAdapter = {
 };
 
 const defaultOptions = {
-  adapter: defaultAdapter
+  adapter: defaultAdapter,
+  onError: e => {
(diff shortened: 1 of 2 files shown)


bjmin:
this changes really helps a lot since we had missed so many error exceptions, which cause us a diffcult debug env.
I just wonder when those can be published

jayphelps:
I'm not confident enough to release any of these yet due to my concerns here: https://github.com/redux-observable/redux-observable/pull/379#issuecomment-349177372

jayphelps:
Closing for now as I believe I've fixed the underlying swallowing issue with rxjs in 5.5.6 https://github.com/ReactiveX/rxjs/pull/3161

PR #385, opened 2017-12-25
typings: make lettable ofType properly narrow action type

### LINKED ISSUE #382: ofType cannot infer that return type is of that type
Suppose you have types for two actions:

```typescript
const PING = "PING";
interface PING {
  type: typeof PING;
}
const ping: () => PING = () => ({
  type: PING,
});

const PONG = "PONG";
interface PONG {
  type: typeof PONG;
}
const pong: () => PONG = () => ({
  type: PONG,
});

type RootAction = PING | PONG;
type RootState = ...;
```

We should now be able to write epics and have proper typing. For example:

```typescript
const epic: Epic<RootAction, RootState> = action$ => 
  action$
    .ofType(PING)
    .map(action => /* The type of action here is still PING | PONG */)
```

Although we come into an issue that `ofType` does not return an observable that is of the type we expect.

We now have to cast the returned action into the type that we expect it to be. I have previously seen that done as below. However, with typescript 2.6 introducing strict function types we can no longer do that either.

```typescript
const epic: Epic<RootAction, RootState> = action$ => 
  action$
    .ofType(PING)
    .map((action: PING)  => /* typescript 2.6 with strict function types doesn't like this. */)
```

One solution to get around this is to use rxjs `filter` instead of using `ofType`. However, by doing this the code becomes a little repetitive and difficult to read. This is especially an issue as the pattern of filtering on action type using redux observable is very common.  

```typescript
const epic: Epic<RootAction, RootState> = action$ => 
  action$
    .filter<RootAction, PING>((action): action is PING => action.type === PING)
    .map(action  => /* action is now of type PING! */)
```

The function ofType is implemented to accept a variadic parameter of keys. However, typescript does not support variadic generic templates (yet). I propose having a number of overloads for different number of parameters.

For example this is the current type definition for `ofType` within the `ActionsObservable` class:
```typescript
ofType(...key: T['type'][]): ActionsObservable<T>;
```

This would become something like this to solve the problem:
```typescript
ofType<T extends Action>(key: T['type']): ActionsObservable<T>;
ofType<T1 extends Action, T2 extends Action>(key1: T1['type'], key2: T2['type']): ActionsObservable<T1 | T2>;
ofType(...key: T['type'][]): ActionsObservable<T>;
```

We can now use ofType and send through the type of the action and get correct type information

```typescript
const epic: Epic<RootAction, RootState> = action$ => 
  action$
    .ofType<PING>(PING)
    .map(action => /* action is now of type PING! */)
```

I would be happy to work on this and add more documentation about using redux observable with typescript but would like some input first. So let me know!

Hotell:

- adds Typescript docs to troubleshooting
- enables strict mode for TS test
- covers proper type narrowing within TS tests

Closes #382

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -51,6 +51,6 @@ export declare function createEpicMiddleware<T extends Action, S, D = any, O ext
 export declare function combineEpics<T extends Action, S, D = any, O extends T = T>(...epics: Epic<T, S, D, O>[]): Epic<T, S, D, O>;
 export declare function combineEpics<E>(...epics: E[]): E;
 
-export declare function ofType<T extends Action, R extends T = T>(...keys: T['type'][]): (source: Observable<T>) => Observable<R>;
+export declare function ofType<T extends Action, R extends T = T, K extends R['type'] = R['type']>(...key: K[]): (source: Observable<T>) => Observable<R>;
 
 export declare const EPIC_END: '@@redux-observable/EPIC_END';
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -125,6 +125,73 @@ const myEpic = action$ => {
 
 This approach essentially returns an empty `Observable` from the epic, which does not cause any downstream actions.
 
+### Typescript: ofType operator won't narrow to proper Observable type
+
+Let's say you have following action types + action creator types:
+
+```ts
+import { Action } from 'redux'
+
+const enum ActionTypes {
+  One = 'ACTION_ONE',
+  Two = 'ACTION_TWO',
+}
+const doOne = (myStr: string): One => ({type: ActionTypes.One, myStr})
+const doTwo = (myBool: boolean): Two => ({type: ActionTypes.Two, myBool})
+
+interface One extends Action {
+  type: ActionTypes.One
+  myStr: string
+}
+interface Two extends Action {
+  type: ActionTypes.Two
+  myBool: boolean
+}
+type Actions = One | Two
+```
+
+When you're using `.ofType` operator for filtering, returned observable won't be correctly narrowed within Type System, because its not capable of doing so yet ( TS 2.6.2 ).
+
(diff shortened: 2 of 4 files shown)


index.d.ts
```diff
@@ -52,6 +52,6 @@ export declare function createEpicMiddleware<T extends Action, S, D = any>(rootE
 export declare function combineEpics<T extends Action, S, D = any>(...epics: Epic<T, S, D>[]): Epic<T, S, D>;
 export declare function combineEpics<E>(...epics: E[]): E;
 
-export declare function ofType<T extends Action>(...keys: T['type'][]): (source: Observable<T>) => Observable<T>;
+export declare function ofType<T extends Action, R extends T, K extends R['type'] = R['type']>(...key: K[]): (source: Observable<T>) => Observable<R>;
```
jayphelps: I believe this would be a breaking change since it now requires two params instead of one. But if we provide a default param for the second one I think that would make it non-breaking and have the same behavior as before wouldn't it? `R extends T = T`

```ts
ofType<T extends Action, R extends T = T, K extends R['type'] = R['type']>
```

jayphelps:
Thank you for this! Let's discuss my breaking change comment a bit.

Also, I kicked the build as it failed for unrelated reasons and should pass now, assuming it would otherwise.

index.d.ts
Hotell: yeah, good catch! my bad thx!

Hotell:
hey @jayphelps sorry it took so long :( too busy  lately
- squashed + rebase + conflicts resolved
- update def to non breaking

thx!

jayphelps:
Thanks again! 🕺 I'm gonna do a release in the next couple days.

jayphelps:
Finally released https://github.com/redux-observable/redux-observable/releases/tag/v0.18.0

Hotell:
🍻

PR #387, opened 2017-12-27
refactor(operator) rewrite the ofType operator

### LINKED ISSUE #386: Refactor the ofType operator
The [ofType](https://github.com/redux-observable/redux-observable/blob/master/src/operators.js#L7) operator has useless `if` statement. As for the filter callback, it could be written with `some` function. Here is the code example: 

```js
const hasType = type => key => {
  return type === key || typeof key === 'function' && type === key.toString();
};

export function ofType(...keys) {
  return source => source::filter(({ type }) => keys.some(hasType(type)));
}
```

iparm:
Closes #386

### DIFF
--- a/src/operators.js
+++ b/src/operators.js
@@ -1,23 +1,9 @@
 import { filter } from 'rxjs/operator/filter';
 
-const keyHasType = (type, key) => {
+const hasType = type => key => {
   return type === key || typeof key === 'function' && type === key.toString();
 };
 
 export function ofType(...keys) {
-  return function ofTypeOperatorFunction(source) {
-    return source::filter(({ type }) => {
-      const len = keys.length;
-      if (len === 1) {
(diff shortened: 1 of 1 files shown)


jayphelps:
Hey there! Thanks for taking the time to make this PR. This is actually a duplicate of #361. btw the extra if statement was a perf optimization since 99.9% of `ofType` usage will only be a single action and, at least at the time, microbenchmarks showed an improvement without the loop. Similar reason for not using `arr.some`. These sorts of optimizations are usually pre-mature in app code, but for libraries it can sometimes be a good idea because they get used in so many different ways, e.g. on extremely low powered devices where nearly every cycle counts, as well as the extra array allocation. I'd still agree with anyones argument that its unlikely to make a notable difference, but then again it doesn't particular hurt since its a simple loop. Thanks again 🕺

PR #403, opened 2018-01-15
chore(rxjs): update rxjs to v6 (WIP)

evertbouw:
BREAKING CHANGE:

RxJS v6 is now required.

Known issues
- [x]  index.d.ts is using two imports from rxjs/internal, I feel like this isn't correct
- [x] test `should throw if you provide a root epic that doesn't return anything` fails. Middleware does throw but test doesn't pick this up


jayphelps:
Thanks @evertbouw for kicking this off! I'm gonna jot down some random notes, mostly for me but feel free to comment.

- [ ] [modify our build process](https://github.com/ReactiveX/rxjs/blob/master/doc/pipeable-operators.md#build-and-treeshaking) to enable tree shaking so that redux-observable doesn't pull in all of rxjs.
- [ ] I also agree we shouldn't need to import from internal, I'll bring up the possibility of rxjs publicly exposing those classes or whether there's a better alternative.
- [ ] I'm going to guess a large percentage of people won't upgrade to v6 quickly. I wonder if we should have a support path for v5 or just bite the bullet? If we bite it
- [ ] Should we do a pre-release with these changes before v6 is final, or wait until then?

razagill:
@jayphelps are there plans of having a pre-release version with rxjs v6 anytime soon?

evertbouw:
I'd like that but imo we should wait for a beta release of RXJS 6. It's a moving target right now and a release might not be too useful?

razagill:
Sounds good

evertbouw:
First beta for Rxjs 6 is out now. Any changes needed before a release is possible?

jayphelps:
Had to resolve some merge conflicts. Also going to follow up with an upgrade of our webpack so that we do tree shaking when we make our builds, otherwise we're pulling everything in rxjs in, though certainly the users tooling might have saved them from us.

Thanks again @evertbouw !! A release is coming today with this after I fix the other stuff.

PR #411, opened 2018-01-24
chore(ci): use LTS and Stable

Shyam-Chen:


jayphelps:
Thanks @Shyam-Chen ! FYI `feat()` commits are for public API features rather than internal things. `chore()` was correct. 😄  It's definitely confusing so don't worry! I fixed it when I merged.

PR #412, opened 2018-01-25
Server side render

Jerry-Hong:

### DIFF
--- a/src/server/wrapClosableEpic.js
+++ b/src/server/wrapClosableEpic.js
@@ -1,0 +1,21 @@
+import { Subject } from 'rxjs/Subject';
+import { ActionsObservable } from '../ActionsObservable';
+
+export const wrapClosableEpic = userEpic => {
+  const inputActions = new Subject();
+  const outputActions = new Subject();
+
+  const rootEpic = (actions, ...rest) => {
+    const closableActions = new ActionsObservable(inputActions);
+    actions.subscribe(inputActions);
+    userEpic(closableActions, ...rest).subscribe(outputActions);
+    return outputActions;
+  };
+
+  rootEpic.close = cb => {
+    inputActions.complete();
+    outputActions.subscribe(null, cb, cb);
+  };
+
+  return rootEpic;
+};
--- a/test/wrapClosableEpic-spec.js
+++ b/test/wrapClosableEpic-spec.js
@@ -1,0 +1,46 @@
+/* globals describe it */
+import { expect } from 'chai';
+import { Subject } from 'rxjs/Subject';
+import { map } from 'rxjs/operator/map';
+import { delay } from 'rxjs/operator/delay';
+import { combineEpics, wrapClosableEpic, ActionsObservable } from '../';
+
+describe('closableEpic', () => {
+  const epic1 = actions =>
+    actions.ofType('ACTION1')::map(action => ({ type: 'DELEGATED1', action }));
+  const epic2 = (actions, store) =>
+    actions
+      .ofType('ACTION2')
+      ::delay(1000)
+      ::map(action => ({ type: 'DELEGATED2', action }));
+
+  const epic = combineEpics(epic1, epic2);
+  const closableEpic = wrapClosableEpic(epic);
+
+  it('closableEpic is a function', () => {
+    expect(closableEpic).to.be.a('function');
+  });
+
+  it('closableEpic with close method', () => {
+    expect(closableEpic.close).to.be.a('function');
+  });
+
+  it('close callback will be hook when all actions sent', () => {
+    const subject = new Subject();
+    const actions = new ActionsObservable(subject);
+
+    const result = closableEpic(actions);
+    const emittedActions = [];
+    result.subscribe(emittedAction => emittedActions.push(emittedAction));
+
+    subject.next({ type: 'ACTION1' });
+    subject.next({ type: 'ACTION2' });
+
+    closableEpic.close(() => {
+      expect(emittedActions).to.deep.equal([
+        { type: 'DELEGATED1', action: { type: 'ACTION1' } },
+        { type: 'DELEGATED2', action: { type: 'ACTION2' } },
+      ]);
+    });
+  });
+});
--- a/index.d.ts
+++ b/index.d.ts
@@ -54,4 +54,6 @@ export declare function combineEpics<E>(...epics: E[]): E;
 
 export declare function ofType<T extends Action>(...keys: T['type'][]): (source: Observable<T>) => Observable<T>;
 
+export declare function wrapClosableEpic<E>(epic: E): E;
+
 export declare const EPIC_END: '@@redux-observable/EPIC_END';
--- a/src/index.js
+++ b/src/index.js
@@ -3,3 +3,4 @@ export { ActionsObservable } from './ActionsObservable';
 export { combineEpics } from './combineEpics';
 export { EPIC_END } from './EPIC_END';
 export { ofType } from './operators';
+export { wrapClosableEpic } from './server/wrapClosableEpic.js';


jayphelps:
@Jerry-Hong awesome! Thanks for kickstarting. Can you elaborate a bit on this approach? How to use it, pros/cons, etc. What do you think about instead of making it per-epic, its instead handled by middleware? redux-saga handles a similar thing by listening for a special END action. The existing EPIC_END could be repurposed for this perhaps.

Jerry-Hong:
@jayphelps  Sorry, I forgot leave this PR info, and my poor English

### usage

server side

```javascript
const route = [
  {
    path: '/',
    redirectTo: '/posts',
    loadData: (dispatch) => {...}
  }
];

app.get('/*', (req, res) => {
    const closableEpic = wrapClosableEpic(rootEpic);
    const epicMiddleware = createEpicMiddleware(closableEpic);
    const store = configureStore({}, epicMiddleware);
   
   // dispatch actions for this url 
   // here is react-router official example
    routes.some(route => {
      const match = matchPath(req.url, route);
      if (match) route.loadData(store.dispatch, match);
      return match;
    });

   closableEpic.close(err => {
     if (err) {  ... }
     // get state and render 
   });
});
```

IMO, in this way we don't need to change any code, it's just a helper function for SSR.

Handle by middleware not per-epic make thing simple. I don't understand why we need handle by per-epic ?

Though takeUntil EPIC_END is good to me,  it maybe is a breaking change who use EPIC_END for hot-reload, I guess.

jayphelps:
> I don't understand why we need handle by per-epic ?

gotcha! per-epic was the wrong term for me to use as I didn't mean every individual epic but rather the root epic. I was referring to the usage of `wrapClosableEpic`

```js
const closableEpic = wrapClosableEpic(rootEpic);
```

I think we can work up a solution where this isn't needed, but instead you can close it either by emitting an END action or maybe just by calling `epicMiddleware.close()` I'm not sure yet though. I need to test a few things.

For anyone interested, here's redux-saga's https://github.com/redux-saga/redux-saga/issues/255

> Though takeUntil EPIC_END is good to me, it maybe is a breaking change who use EPIC_END for hot-reload, I guess.

That's definitely true. If does end up being the right thing to do I personally think it will be okay. We're already having some breaking changes coming for v1 like [state$ second argument](#410) so now would be a good time if we do decide that.

***

Are there benefits to the `wrapClosableEpic` approach over having it built into the middleware?

jayphelps:
hey @Jerry-Hong just double checking you saw this:

> Are there benefits to the wrapClosableEpic approach over having it built into the middleware?

😃  no worries if you're busy

Jerry-Hong:
@jayphelps 
Built into the middleware made API more simple for user. I agree do it if we can!

chiel:
This looks exactly like something I need. The only thing I see that I might change, and I'm nitpicking here, is to call the `closableEpic.close` function `closableEpic.onClose` instead, to make it clear you can attach a handler to it. `closableEpic.close` makes it seem to me like that forces all running epics to be cancelled.

chiel:
Out of curiosity - what's currently blocking this from being merged? Are there outstanding issues that need to be resolved or something that still needs to be discussed? Not trying to be pushy, but this is definitely something I could use on the short-ish term. :)

Jerry-Hong:
@chiel we discuss about this feature should built into middleware or just a helper function for v1.0.0. There is not definitely way now.

@jayphelps Is it possible merge this for v0.19.0? And we create another issue for v1.0.0 discussing?

chiel:
Well, I don't mean to push you guys into rushing this one, I could always copy the closableEpic stuff for now and use that :)

jayphelps:
Sorry for the delay folks, I [just started a new gig](https://twitter.com/_jayphelps/status/962367319576948736) and should have a bit more time to devout to my OSS stuff, including this project.

The holdup on this particular PR is that I'm currently convinced this should be part of the middleware API itself rather than having people wrap their epics.

I have not yet worked on an app, or helped others with their's, that required SSR redux-observable so I don't have enough experience to concretely know the desired behavior and potential issues. Since epics are where side effects happen, potentially infinitely running ones, the experimentation I've done previously around APIs left me thinking it was highly variable depending on a given developers expectations. I've researched how redux-saga users approach the use case and seen similar confusion and conflicting expectations.

Still, providing _something_ is probably better than nothing as long as its documented well. Is @Jerry-Hong and anyone else interested in jumping on a Google Hangout discussion around this to hash out a potential middleware-based API? I have a bunch of questions for people who truly have the SSR use case already.

***

@chiel 

> call the closableEpic.close function closableEpic.onClose instead, to make it clear you can attach a handler to it. closableEpic.close makes it seem to me like that forces all running epics to be cancelled.

It actually does trigger the cancel closing processing, rather than just register a listener. The `inputActions.complete()` call inside is what kicks it off.

chiel:
@jayphelps alright, in that case it makes total sense. I'll just copy over the `closableEpic` stuff for now in a project where I need it, and then when there is a better (read: alternative) solution I'll amend my code.

I would hop into a hangout, but I'm from Amsterdam so I think the time zone difference is a bit of a dealbreaker there. Additionally, I've only just started using redux-observable + redux so I'd probably be more of a hinder than a help in such a discussion. I'll definitely keep following this PR though, excited to see where it goes.

Jerry-Hong:
@jayphelps My Google Email is s6323859@gmail.com

I'm from Taiwan, time zone is +8. Maybe we can discuss at my Sunday 1:00~2:00 a.m.(California Saturday 9:00 - 10:00 a.m.).

chiel:
Any progress on this, by any chance? :)

hitmands:
#539 gives a first basic implementation of #502

morsmodr:
@jayphelps I hope I am not too late to this conversation. I work on a large enterprise application and am looking into making it SSR. The problem that I have might be the one you are looking to hear from for across PRs [https://github.com/redux-observable/redux-observable/pull/412](412), [https://github.com/redux-observable/redux-observable/issues/502](502), [https://github.com/redux-observable/redux-observable/pull/539](539) and here :)

The application will use getServerSideProps as it isn't an entirely static page. The way epics have been implemented in our application is that they are infinite in nature. Think of it like a side effect manager across multiple reducers within 1 mega store for the application. Various sections of the page have their own reducers and user interaction in 1 section of the page can result in re-computation of data shown in other sections (and thereby the need to update other reducers) of the website. The application itself is a SPA with 2 routes to it, and decisions in either page need to update info on the other route as well.

I tried the solution of $shutdown given in the stackblitz example and as a result my NextJS app never loads at all because we don't use RO for ajax calls, so there is no unsubsribe, end or complete in sight for the epics in question.

What would be a good way to tackle such an implementation of RO with SSR?
  I have come across next-redux-wrapper and next-redux-observable but not sure whether that is the way forward. Plus adding 2 libraries is making me think about it, as I prefer to have fewer dependencies in general. Moving over to redux-saga is another idea that I have seen floating around but having known RXJS, RO was super easy for me and members of my team to pick-up and get going.

src/server/wrapClosableEpic.js
```diff
@@ -0,0 +1,21 @@
...
+import { ActionsObservable } from '../ActionsObservable';
+
+export const wrapClosableEpic = userEpic => {
+  const inputActions = new Subject();
+  const outputActions = new Subject();
+
+  const rootEpic = (actions, ...rest) => {
+    const closableActions = new ActionsObservable(inputActions);
+    actions.subscribe(inputActions);
+    userEpic(closableActions, ...rest).subscribe(outputActions);
```
getkey, line 11:
```suggestion
    actions.subscribe(inputActions);
    userEpic(inputActions.asObservable(), ...rest).subscribe(outputActions);
```

test/wrapClosableEpic-spec.js
```diff
@@ -0,0 +1,46 @@
...
+
+  it('closableEpic with close method', () => {
+    expect(closableEpic.close).to.be.a('function');
+  });
+
+  it('close callback will be hook when all actions sent', () => {
+    const subject = new Subject();
+    const actions = new ActionsObservable(subject);
+
+    const result = closableEpic(actions);
```
getkey, line 32:
```suggestion
    const result = closableEpic(subject.asObservable());
```

getkey (changes requested):
We've been using this code in production for years now, works great.
However, `ActionsObservable` [has been deprecated in v2](https://github.com/redux-observable/redux-observable/blob/master/CHANGELOG.md#200-alpha0-2019-11-14), so this PR no longer works.
The fix is simple thankfully 🙂

evertbouw:

PR #428, opened 2018-02-14
Fix(scripts): Use rimraf to fix script on Windows

evertbouw:


jayphelps:
Thanks @evertbouw! 👏  I changed the commit message to be a `chore()` as `fix()` is meant to be bugfixes that would affect consumers of the library, rather than internal build process fixes. That's definitely not obvious, so no worries! The reason is that `fix()` will show up in the automated CHANGELOG but `chore()` will not.

PR #435, opened 2018-02-22
Add prettier

### LINKED ISSUE #433: Add prettier
Hey!

I think `prettier` is such a great tool and having a tool to format your code makes it easier to everyone to just focus on the code and not how to satisfy `eslint` how certain parts should be formatted.

I could add prettier and make a lint rule and also use the eslint config to disable the rules that are already there that demand styling.

I'm somehow against having `prettier ` as an `eslint` rule, I'd just add `prettier` as one of the first steps of `npm run lint`

I think prettier can handle the function bind syntax currently, at least it doesn't seem to break it.

cilice:
I added prettier and ran it on all possible files that prettier can run except for json files.

If it's too much and you want to keep the format on certain files, just comment away :)

Fixes #433


CHANGELOG.md
```diff
@@ -1,32 +1,28 @@
 <a name="0.18.0"></a>
-# [0.18.0](https://github.com/redux-observable/redux-observable/compare/v0.17.0...v0.18.0) (2018-02-07)
 
+# [0.18.0](https://github.com/redux-observable/redux-observable/compare/v0.17.0...v0.18.0) (2018-02-07)
```
jayphelps: We'll need to ignore this file since it is autogenerated and would always need reformatting for every release. https://prettier.io/docs/en/ignore.html

SUMMARY.md
```diff
@@ -3,28 +3,32 @@
...
+* [Troubleshooting](docs/Troubleshooting.md)
 
 ## [API Reference](docs/api/SUMMARY.md)
-  * [createEpicMiddleware](docs/api/createEpicMiddleware.md)
-  * [combineEpics](docs/api/combineEpics.md)
-  * [EpicMiddleware](docs/api/EpicMiddleware.md)
+
+* [createEpicMiddleware](docs/api/createEpicMiddleware.md)
+* [combineEpics](docs/api/combineEpics.md)
+* [EpicMiddleware](docs/api/EpicMiddleware.md)
```
jayphelps: I need to double check that this doesn't actually change the git-book generated output as I have a vague recollection that the indention mattered.

CHANGELOG.md
cilice: Added to the ignore and amended :)

docs/basics/Epics.md
```diff
@@ -25,32 +26,29 @@ If you let an incoming action pass through, it will create an infinite loop:
...
 
-
 Let's start with a simple Epic example:
 
 ```js
 const pingEpic = action$ =>
-  action$.filter(action => action.type === 'PING')
-    .mapTo({ type: 'PONG' });
-    
+  action$.filter(action => action.type === 'PING').mapTo({ type: 'PONG' });
```
jayphelps: I'd like app epics to be consistent even if they don't have a lot of operators applied. The other cases where there were 3+ prettier decided to put them all on newlines like

```js
action$
  .filter(action => action.type === 'PING')
  .mapTo({ type: 'PONG' });
```

Not sure if this is easy to do.

docs/basics/Epics.md
```diff
@@ -134,12 +134,14 @@ const fetchUserFulfilled = payload => ({ type: FETCH_USER_FULFILLED, payload });
 
 // epic
 const fetchUserEpic = action$ =>
-  action$.ofType(FETCH_USER)
+  action$
+    .ofType(FETCH_USER)
     .mergeMap(action =>
-      ajax.getJSON(`https://api.github.com/users/${action.payload}`)
+      ajax
+        .getJSON(`https://api.github.com/users/${action.payload}`)
```
jayphelps: I'm not sure I like this opinion. I didn't care as much on the `ofType` examples but this one ends up being weird and not something I've seen anyone ever do. 🤔

docs/basics/Epics.md
```diff
@@ -220,21 +224,16 @@ Finally, redux-observable provides a utility called [`combineEpics()`](../api/co
...
 
 ```js
 import { merge } from 'rxjs/observable/merge';
 
-const rootEpic = (action$, store) => merge(
-  pingEpic(action$, store),
-  fetchUserEpic(action$, store)
-);
+const rootEpic = (action$, store) =>
+  merge(pingEpic(action$, store), fetchUserEpic(action$, store));
```
jayphelps: While I don't mind the new variant, I imagine the prior version is easier for newcomers to grok. Not a big deal, curious what others think.

docs/recipes/Cancellation.md
```diff
@@ -8,25 +8,24 @@ This can be done with the [`.takeUntil()`](http://reactivex.io/rxjs/class/es6/Ob
...
-      ajax.getJSON(`/api/users/${action.payload}`)
-        .map(response => fetchUserFulfilled(response))
-        .takeUntil(action$.ofType(FETCH_USER_CANCELLED))
-    );
+  action$.ofType(FETCH_USER).mergeMap(action =>
+    ajax
+      .getJSON(`/api/users/${action.payload}`)
+      .map(response => fetchUserFulfilled(response))
+      .takeUntil(action$.ofType(FETCH_USER_CANCELLED))
+  );
```
jayphelps: this one (and the others like it) are yucky IMO 😢

I totally get why it would default to it, and in many traditional cases I would prolly like it. I just think in the case of epics they become more maintainable if the formatting is always consistent. e.g. if you had added a debounce above the mergeMap it wouldn't have put them all on one line

index.d.ts
```diff
@@ -11,16 +11,57 @@ export declare class ActionsObservable<T extends Action> extends Observable<T> {
...
    * provide safe typing for up to 6 items, following by a scheduler.
    */
-  static of<T extends Action>(item1: T, scheduler?: Scheduler): ActionsObservable<T>;
-  static of<T extends Action>(item1: T, item2: T, scheduler?: Scheduler): ActionsObservable<T>;
-  static of<T extends Action>(item1: T, item2: T, item3: T, scheduler?: Scheduler): ActionsObservable<T>;
-  static of<T extends Action>(item1: T, item2: T, item3: T, item4: T, scheduler?: Scheduler): ActionsObservable<T>;
-  static of<T extends Action>(item1: T, item2: T, item3: T, item4: T, item5: T, scheduler?: Scheduler): ActionsObservable<T>;
-  static of<T extends Action>(item1: T, item2: T, item3: T, item4: T, item5: T, item6: T, scheduler?: Scheduler): ActionsObservable<T>;
-  static of<T extends Action>(...array: Array<T | Scheduler>): ActionsObservable<T>;
+  static of<T extends Action>(
```
jayphelps: ehhh 👎 but I totally get it. It's one of those "in this specific case I disagree, but in most cases I would agree"

cilice:
I also somehow went for the all-in approach with trying to give all markdown files to prettier. 

If it's easier to not enforce style on the markdown files and also leave the ts definition file as it was, I wouldn't mind it. It was just an option with the aim start a discussion around that.

jayphelps:
@cilice Right on. To be clear, I use prettier on apps I work on, so I love it. I just don't agree with all the opinions it has.

Does anyone else have opinions here? It's not a huge deal to just bend over and let prettier tell us what to do even if we don't like it but I think we'd want prettier to run automatically via something like a [pre-commit hook](https://prettier.io/docs/en/precommit.html)

docs/basics/Epics.md
ohjames: The old version is prettier and easier to understand but the hassle "prettier" saves is worth accepting a few weird choices for.

evertbouw:
Prettier 1.13 as been released and now functions like pipe and compose will always have their params multilined. It doesn't work on stuff like `combineEpics`, `concat`, or `merge` but I think I can live with that. Can you update to the latest version and run it again?

index.d.ts
JLarky: yeah, adding prettierignore block can be good here

Issue #437, opened 2018-02-23
Could typeOf narrowing work without explicit generics in TypeScript 2.8?


eemelisoderholm:
While #385 did provide a solution for maintaining action type safety through `ofType<Action, ActionType>(action)` filtering, it's a bit inconvenient to use and I'm wondering if predefined conditional types in TS 2.8 (https://github.com/Microsoft/TypeScript/pull/21847) would make it possible to narrow down the type from one big union type, without passing in the explicit type.

I'd imagine most people want to infer action types from action creators and not write explicit type definitions, because of how verbose it gets with lots of actions. This is pretty viable now (see https://medium.com/@martin_hotell/improved-redux-type-safety-with-typescript-2-8-2c11a8062575 for example).

At least from the reducer's point of view, it's fine to have one big action union type, which can be derived from all action creators with `ReturnType<typeof actionCreator>`. Narrowing works just as you'd expect, just matching type with string enum or constant.

I kinda expected it to work the same way in my epics as well, but as the troubleshooting manual says, I need to pass in the exact action type to the generic typeOf for the narrowing to work.

For now I'm doing something like this

```typescript
// src/state/example/example.epics.ts
import { Epic, ofType, combineEpics } from 'redux-observable'
import { switchMap } from 'rxjs/operators/switchMap'
import { ActionAny, RootState } from '../index'
import * as actions from './example.actions'

const exampleFetchAllEpic: Epic<ActionAny, RootState> = (action$, store) =>
  action$.pipe(
    ofType<ActionAny, ReturnType<typeof actions.exampleActionFetchAll>>(
      actions.ExampleActionType.FETCH_ALL
    ),
    switchMap(action => {
      console.log(action.payload.foo)
    })
  )

export const epics = combineEpics(
  exampleFetchAllEpic
)
```

Now I don't quite understand why the narrowing requires this pattern and it's not a huge issue, so sorry if it's simply not viable, but I feel like the action type should be implicit since ActionAny is already union of all the return types.

Also TS 2.8 isn't stable release yet so I don't expect it to be supported either, but just wanted to bring this up in advance.

Edit: After going through other related issues, I realised the conditional types in 2.8 may not solve the issue here. Still, it feels redundant to declare the exact action type and check the type field, so I'm curious to what needs to happen for the TS compiler to understand the narrowing, or if anyone has workarounds for it.

jayphelps:
[...]

Issue #417, opened 2018-02-02
Types not Narrowed through ofType [TypeScript]


cjol:
Using `action$.ofType(...)` does not narrow the type of the ActionsObservable. Instead, the resulting Observable has the same type as `action$`, which means any subsequent operators don't have access to much useful type information. For example, this gives a type error on the access to `action.firstProp`: 

```
export const pingEpic: Epic<MyActions, StateType> = (action$, store) =>
  action$
    .ofType("PING")
    .delay(1000)
    .map((action): ActionTwo => ({
      type: "PONG",
      secondProp: action.firstProp + " 2"
    }));
```

Demo: https://stackblitz.com/edit/typescript-f4qzhx?file=ping-pong.ts

Note that in the demo above you can get rid of the type error by adding a type annotation to the map (e.g. `.map((action:ActionOne):ActionTwo => ({...})`. This only works because strict mode is not enabled on StackBlitz.

Instead, a new type definition is needed for ofType:

```diff
-   ofType(...key: T["type"][]): ActionsObservable<T>;
+   ofType<S extends T>(...key: S["type"][]): ActionsObservable<S>;
```

Making this change on my local copy of the .d.ts file allows me to do this, and have access to all the lovely type information:

```
export const pingEpic: Epic<MyActions, StateType> = (action$, store) =>
  action$
    .ofType<ActionOne>("PING")
    .delay(1000)
    .map((action): ActionTwo => ({
      type: "PONG",
      secondProp: action.firstProp + " 2"
    }));
```

What do you think?

jayphelps:
@cjol Can you take a look at the latest code in master? https://github.com/redux-observable/redux-observable/commit/63b2accd0a4ff3d8665a0e66a653ad819a4f62ec

```ts
ofType<R extends T = T>(...key: R['type'][]): ActionsObservable<R>;
```

I believe this basically the same thing you're asking for right? It hasn't made it into a release yet, but I expect one this weekend.

Let me know if I'm confused, and this is different 😄 

***

FYI for everyone coming here later: the latest word direct from the TypeScript team was that it was not possible to automatically narrow the union without some sort of explicit type param hint. So for now, to narrow you must explicitly pass the correct type param.

* https://github.com/Microsoft/TypeScript/issues/17915
* https://stackoverflow.com/questions/46312206/narrowing-a-return-type-from-a-generic-discriminated-union-in-typescript/46313086#46313086

cjol:
[...]

PR #459, opened 2018-03-27
Automatically narrow Action type in the ofType operator

zsparal:
BREAKING CHANGE: The TypeScript definitions now require TS 2.8+ and `ofType` operations in existing codebases will also stop working (they will fail with a compilation error in case they specified the action type as a generic parameter)

With the new conditional types in TS 2.8 it's now possible to correctly narrow action types in `ofType` (and there are even some nice standard library types to do so).

The only downside of the current implementation is that the lettable version of the operator cannot restrict the the key types to the types used by the actions themselves. Compare the definition of the chained version:

```ts
ofType<K extends T['type']>(...)
```

With the lettable one:

```ts
function ofType<T extends Action, K extends string>(...)
```

Unfortunately in the second case, changing the constraint to `K extends T['type']` breaks the action type inference needed for the automatic narrowing.

As far as I can see this is unavoidable but it doesn't really cause any problems except for slightly worse error messages. If you pass a type that is not compatible with the action:

1. In the chaining case: you get an error telling you that the parameter is not assignable to the action type
2. In the lettable case: the narrowed action type will be `never` so you essentially won't be able to use it for anything useful

### DIFF
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -121,53 +121,6 @@ const myEpic = action$ => {
 
 This approach essentially returns an empty `Observable` from the epic, which does not cause any downstream actions.
 
-### Typescript: ofType operator won't narrow to proper Observable type
-
-Let's say you have following action types + action creator types:
-
-```ts
-import { Action } from 'redux'
-
-const enum ActionTypes {
-  One = 'ACTION_ONE',
-  Two = 'ACTION_TWO',
-}
-const doOne = (myStr: string): One => ({ type: ActionTypes.One, myStr })
-const doTwo = (myBool: boolean): Two => ({ type: ActionTypes.Two, myBool })
-
-interface One extends Action {
-  type: ActionTypes.One
-  myStr: string
-}
-interface Two extends Action {
-  type: ActionTypes.Two
-  myBool: boolean
-}
-type Actions = One | Two
-```
-
-When you're using `ofType` operator for filtering, returned observable won't be correctly narrowed within Type System, because its not capable of doing so yet ( TS 2.6.2 ).
-
-To fix this, you need to explicitly set the generic type, so Typescript understands your intent, and narrows your action stream correctly:
-
-```ts
-// This will let action be `Actions` type, which is wrong
-const epic = (action$: ActionsObservable<Actions>) =>
-  action$.pipe(
-    ofType(ActionTypes.One),
-    // action is still type Actions instead of One
-    map((action) => {...})
-  );
-
-// Explicitly set generics fixes the issue
-const epic = (action$: ActionsObservable<Actions>) =>
-  action$.pipe(
-    ofType<One>(ActionTypes.One),
-    // action is correctly narrowed to One
-    map((action) => {...})
-  );
-```
-
 * * *
 
 ## Something else doesn’t work
--- a/index.d.ts
+++ b/index.d.ts
@@ -23,7 +23,7 @@ export declare class ActionsObservable<T extends Action> extends Observable<T> {
   constructor(input$: Observable<T>);
   lift<R extends Action>(operator: Operator<T, R>): ActionsObservable<R>;
   lift<R>(operator: Operator<T, R>): Observable<R>;
-  ofType<R extends T = T>(...key: R['type'][]): ActionsObservable<R>;
+  ofType<K extends T['type']>(...key: K[]): ActionsObservable<Extract<T, { type: K }>>;
 }
 
 export declare class StateObservable<S> extends Observable<S> {
@@ -51,4 +51,4 @@ export declare function combineEpics<T extends Action, O extends T = T, S = void
 export declare function combineEpics<E>(...epics: E[]): E;
 export declare function combineEpics(...epics: any[]): any;
 
-export declare function ofType<T extends Action, R extends T = T, K extends R['type'] = R['type']>(...key: K[]): (source: Observable<T>) => Observable<R>;
+export declare function ofType<T extends Action, K extends string>(...key: K[]): (source: Observable<T>) => Observable<Extract<T, { type: K }>>;
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     "rimraf": "^2.5.4",
     "rxjs": "^6.0.0",
     "sinon": "^4.5.0",
-    "typescript": "^2.1.4",
+    "typescript": "^2.8.1",
     "webpack": "^4.5.0",
     "webpack-cli": "^2.0.13",
     "webpack-rxjs-externals": "~2.0.0"
--- a/test/typings.ts
+++ b/test/typings.ts
@@ -100,7 +100,7 @@ interface Epic9_Output {
 
 const epic9_1: Epic<FluxStandardAction, Epic9_Output, State, Dependencies> = (action$, state$, dependencies) =>
   action$.pipe(
-    ofType<FluxStandardAction, Epic9_Input>('NINTH'),
+    ofType('NINTH'),
     map(({ type, payload }) => ({
       type: 'ninth' as 'ninth',
       payload: dependencies.func("ninth-" + payload),
@@ -109,7 +109,7 @@ const epic9_1: Epic<FluxStandardAction, Epic9_Output, State, Dependencies> = (ac
 
 const epic9_2 = (action$: ActionsObservable<FluxStandardAction>, state$: StateObservable<void>, dependencies: Dependencies) =>
   action$.pipe(
-    ofType<FluxStandardAction, Epic9_Input>('NINTH'),
+    ofType('NINTH'),
     map(({ type, payload }) => ({
       type: 'ninth',
       payload: dependencies.func("ninth-" + payload),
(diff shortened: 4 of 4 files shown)


jayphelps:
Super thanks for doing this! It's not 100% clear what this improves, can you elaborate with some specific examples of before/after perhaps? It's likely we'll need to hold off on this until a large percentage of the community has moved to this new version of TS since this could be a major pain point for people. Any thoughts on that as well?

Gustorn:
Sure. Just for reference, there has been some issues related to this already: #382, #417, #437. The solution to these issues was the introduction of a generic parameter on `ofType` that allows you to explicitly specify its return type. This PR allows TS to automatically infer this return type in a type-safe manner. Let's see an example:

```ts
const enum ActionTypes {
  One = "One",
  Two = "Two",
  Three = "Three"
}

interface One { type: ActionTypes.One, myStr: string; }
interface Two { type: ActionTypes.Two, myBool: boolean; }
interface Three { type: ActionTypes.Three, myBool: boolean; }
type Actions = One | Two | Three;
```

We can see that calling `ofType` on `ActionObservable<Actions>` and passing the correct `ActionType` correctly narrows the action type (note that we can access `myStr` in the `map`):

image



Whereas previously this didn't work:

image

The workaround was to use an explicit type annotation to specify the return value:

image

While this works, there's a downside (as described in #437): with TS 2.8, the explicit definition of Action interfaces has become a bit redundant since you can just use the newly introduced `ReturnType` to infer them from action creators. This reduces the TS/Redux boilerplate quite a bit but makes `ofType` operations really noisy (see the example in #437).

Just as a quick note, the automatic narrowing also works for multiple action types:

image

 A downside is that people who do not use union types to define their actions would lose the ability to narrow the return type of `ofType`. I can look into maybe adding an overload to also support the old version although I don't think it will be possible because of ambiguity between the two overloads.

As for getting this merged: I'm completely fine with it being delayed (or even not getting merged at all). I mainly sent this PR as a proof of concept and because I needed it for our project anyway. And since type definitions are really easy to override locally it might come in handy for other people as well.

ksaldana1:
@Gustorn Thanks for putting this up! I've been experimenting with a bunch of our redux stuff using some of the new TS 2.8 features as well. That narrowed action type with `Extract` has some cool use cases, as well as programmatic union types inferred from action creators. I actually ran into your [gist](https://gist.github.com/Gustorn/50071757fa8e88195d215acd877117ae) at some point which gave me some ideas, so thanks!

I'm a big fan of this approach, but I do understand that it's a bit much to expect all consumers of this library to be on the brand new version of TS for boilerplate niceties. I'm in a similar position where I don't need this merged immediately--we are already forked using state$ stream as second argument of epics.

jayphelps:
Does anyone have any insight into whether TS 2.8 has enough adoption yet? I know a huge amount of the community pushing forward is trickled down by Angular users, and [they have not yet updated](https://github.com/angular/angular/issues/23779) so that's probably why I haven't seen it adopted in the wild as much yet.

ksaldana1:
Most of the large open-source TS libraries I use (rxjs, mobx, etc) seem to have made the jump, but it's hard to know if that's trickled down to to those users. It would be really cool if you could see npm downloads by version, but that does not seem to be a thing.

2.8 was a big feature release with conditional types and inference which allow us to do things like this PR. 3.0 is slated for July, which will include additional features (`unknown` top type, better handling of rest args/tuples, named generic arguments), but I don't see any features that would be immediately relevant to this library (outside of cleaning up the d.ts file, maybe). 

I'm pretty surprised Angular has fallen behind two versions (TS@latest is 2.9.2)--seems like they'd be a pretty good benchmark to follow for adoption.

ksaldana1:
The [PR](https://github.com/angular/angular/pull/24652) for TS 2.9 seems to be in a pretty late stage--Miško is putting that PR together, so I'd expect it to make it into the main branch pretty soon.

ksaldana1:
[Angular 6.1](https://blog.angular.io/angular-v6-1-now-available-typescript-2-9-scroll-positioning-and-more-9f1c03007bb6) was released recently, which includes support up to TS 2.9 now. I think it'd be great to get this feature added for us TS users. Having to provide the generic for `ofType` isn't the worst thing in the world, but it feels like a cast which is obviously less ergonomic than the solution proposed here.

jayphelps:
@madshargreave ultimately I'm worried about the error messages produced by the lettable version of `ofType` with this solution.

> The only downside of the current implementation is that the lettable version of the operator cannot restrict the the key types to the types used by the actions themselves.

> In the lettable case: the narrowed action type will be never so you essentially won't be able to use it for anything useful

I was playing with these changes and even though I had read this before, I forgot and was pretty confused when I saw `The type 'never' has no property 'payload' and no string index signature`. Made worse is that this error is not when you try and filter your actions with an unknown action type, but rather when you later try to use that filtered action.

```ts
const epic: Epic<{ type: 'FIRST', payload: string }> = action$ =>
  action$.pipe(
    ofType('DOES_NOT_EXIST'), // no error here, even though ideally it would be
    map(({ payload }) => ({   // error is here, not above
      type: 'SECOND',
      payload
    }))
  );
```

I've played around with various attempts, and also kept an eye on what [ngrx/effects is thinking about doing](https://github.com/ngrx/platform/pull/1183) but they too have the same issue.

I'm not yet convinced this is worth the confusion it creates, but I do welcome anyone who has thoughts on why my fears are unfounded?

evertbouw:
Since this is now merged in ngrx/effects I revisited it. I really don't like how it narrows to `never`. Aside from the problem highlighted above, if you use two actions where one is found but not the other, TS doesn't complain but it will cause runtime errors:

```js
const epic: Epic<{ type: 'FIRST', payload: string }> = action$ =>
  action$.pipe(
    ofType('FIRST', 'DOES_NOT_EXIST'), // no error here, even though ideally it would be
    map(({ payload }) => ({ // possible runtime error here
      type: 'SECOND',
      payload
    }))
  );
```

ksaldana1:
@evertbouw I do not believe that code snippet is valid with the current type definitions.
```typescript
export declare interface Epic<Input extends Action = any, Output extends Input = Input, State = any, Dependencies = any> {
  (action$: ActionsObservable<Input>, state$: StateObservable<State>, dependencies: Dependencies): Observable<Output>;
}
```
Because of the second generic default argument `Output extends Input = Input`  your code will break with the error `Type 'string' is not assignable to type '"FIRST"'`.

So let's fix that example:
```typescript
const epic: Epic<
  { type: 'FIRST'; payload: string } | { type: 'SECOND'; payload: string }
> = action$ =>
  action$.pipe(
    ofType('FIRST'),
    map(({ payload }) => ({
      type: 'SECOND',
      payload,
    }))
  );
```
If you run through this code you'll see that you are receiving a similar error: ` Type 'string' is not assignable to type '"SECOND"'`. 

This error is actually due to the `type` argument in the return type of your map operator. To make this code compile you'll need to add a type assertion.

```typescript
const epic: Epic<
  { type: 'FIRST'; payload: string } | { type: 'SECOND'; payload: string }
> = action$ =>
  action$.pipe(
    ofType('FIRST'),
    map(({ payload }) => ({
      type: 'SECOND' as 'SECOND',
      payload,
    }))
  );
```

This finally compiles correctly. This is a behavior that comes up often when dealing with string discriminants in TypeScript. See [this blog post](https://blog.mariusschulz.com/2017/02/04/typescript-2-1-literal-type-widening) for some more examples.

I like the behavior in this PR, but I can agree with the errors being unhelpful with out a lot of upfront assumptions about your TS + Redux setup (e.g. using string enums for action types, always using typesafe action creators to avoid issues such as above, etc.). 

One of the nice things about pipeable operators is that it is trivial to wrap your own operators if you need stricter typings, better ergonomics, etc. As this is still a very generic Redux utility library with little other opinions about your system, I'm fine with the decision to wait a bit on this.

ksaldana1:
I will say that I think guiding people towards best practice is not a bad thing for a library to do. Your `DOES_NOT_EXIST` action "typo" is circumvented by using a string enum for the ActionTypes. It has lots of other benefits from a type system level, but I think guiding people towards that in documentation would probably be good.

I do still think the types in this PR are a more accurate representation of how the types flow. A typo in a generic cast is going to give me the same level of confusion as dealing with a `never` type. Perhaps this is just my personal experience with TS, so I can't speak for everyone.

I see a slight downside in avoidable edge cases (like the typo example) with a great deal of upside (leveraging a powerful static type system to let code drive type inference).

ducin:
I'm experiencing the same issues with:
`"rxjs": "^6.3.3"`, `"redux-observable": "^1.0.0"`. Is there a plan to ship the fix somehow soon?

RayLuxembourg:
Please update this issue.

jayphelps:
No changes. Previous discussion is the status quo, unless TypeScript has released the ability to narrow these typings without resorting to workarounds and without confusing type error messages when not in the happy path.

https://github.com/redux-observable/redux-observable/pull/459#issuecomment-419183556

ksaldana1:
@jayphelps @Gustorn I believe with [propagated generics in higher order functions](https://github.com/Microsoft/TypeScript/pull/30215) released in TS 3.4, the following will narrow correctly:
```typescript
export declare function ofType<T extends Action, K extends T['type']>(
  ...key: K[]
): (source: Observable<T>) => Observable<Extract<T, { type: K }>>;
````

It no longer loses inference and falls back to `{}`. I know it will probably be a while before we can safely assume users are on 3.4, but I think it's good to keep in the backlog for future cleanup.

future-pirate-king:
Any update on this ?

evertbouw:

PR #465, opened 2018-04-05
fix(typings): Update type of Epic to add state$

cjol:
I had to add this to use it in my own project, so I figured I would share.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -39,7 +39,7 @@ export declare class StateObservable<S> extends Observable<S> {
 
 
 export declare interface Epic<T extends Action, S, D = any, O extends T = T> {
-  (action$: ActionsObservable<T>, store: StateObservable<S>, dependencies: D): Observable<O>;
+  (action$: ActionsObservable<T>, state$: StateObservable<S>, dependencies: D): Observable<O>;
 }
 
 export interface EpicMiddleware<T extends Action, S, D = any, O extends T = T> extends Middleware {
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -26,8 +26,20 @@ export declare class ActionsObservable<T extends Action> extends Observable<T> {
...
+  value: S
+
+  // deprecated
+  getState(): S
+  dispatch: Store<S>["dispatch"]
+}
+
+
 export declare interface Epic<T extends Action, S, D = any, O extends T = T> {
-  (action$: ActionsObservable<T>, store: MiddlewareAPI<S>, dependencies: D): Observable<O>;
```
jayphelps: Can you rename the parameter to state$?

jayphelps:
Thank you!! Just the minor param rename and this looks good.

index.d.ts
jayphelps: I tried making this change myself but it looks like you don't have the option checked to "allow edits from maintainers". https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

cjol: Hm, that's strange - I think that option is enabled so I wonder if there is another issue. First pull request so I'm on unfamiliar ground!

Anyway, I think this is now fixed.

ajcrites: This change doesn't remove the import of `MiddlewareAPI` which is no longer used.

PR #471, opened 2018-04-09
feat(build): use rollup to build

Shyam-Chen:


rollup.config.js
```diff
@@ -0,0 +1,50 @@
...
+      presets: [
+        [
+          'env', {
+            modules: false,
+            targets: {
+              browsers: ['last 2 versions']
+            }
+          }
+        ],
+        'stage-3'
```
evertbouw: I don't think we need to use stage-3, the only feature from it that we use is transform-object-rest-spread.

rollup.config.js
```diff
@@ -0,0 +1,50 @@
...
+      ],
+      plugins: [
+        'external-helpers',
+      ],
+    }),
+    isProd && uglify(),
+  ].filter(Boolean),
+  external: Object.keys(pkg.dependencies),
```
evertbouw: There are no dependencies, so this array will be empty. This causes an unresolved dependencies warning. Replacing this with peerDependencies doesn't fix the warning for the submodule import `rxjs/operators`. I think a static array like `['rxjs', 'rxjs/operators', 'redux']` will be fine

rollup.config.js
```diff
@@ -0,0 +1,50 @@
...
+    {
+      file: isProd ? './lib/redux-observable.esm.min.js' : './lib/redux-observable.esm.js',
+      format: 'es',
+      sourcemap: true,
+    },
+    {
+      file: isProd ? './dist/redux-observable.umd.min.js' : './dist/redux-observable.umd.js',
+      format: 'umd',
+      name: 'ReduxObservable',
+      sourcemap: true,
```
evertbouw: You can add globals here to suppress the warning `(!) Missing global variable names`

package.json
```diff
@@ -2,17 +2,14 @@
   "name": "redux-observable",
   "version": "1.0.0-alpha.1",
   "description": "RxJS based middleware for Redux. Compose and cancel async actions and more.",
-  "module": "lib/esm/index.js",
-  "main": "lib/cjs/index.js",
+  "main": "./lib/redux-observable.cjs.js",
+  "module": "./lib/redux-observable.esm.js",
+  "browser": "./dist/redux-observable.umd.js",
```
jayphelps: My initial reaction is that I'd prefer not to change these locations. Lmk if there's a good argument for it? Also curious: where the rest of the flat files, other than the entry points, live if everything is built into the same directory?

jayphelps:
@Shyam-Chen Thanks for spearheading this! Left some comments, lmk if I'm mistaken on any of them or if you have counter thoughts. Also it's totally not obvious but this should be a chore() not a feat() as it's not an API code feature but rather just an internal build process change that _should_ be completely transparent to people if we don't change the filenames (one of my comments above)

For everyone's reference, can you elaborate on the pros and cons of this?

package.json
Shyam-Chen: Change lib to dist?

Shyam-Chen:
UMD Example: https://codepen.io/Shyam-Chen/pen/PRVGPm

rollup.config.js
```diff
@@ -0,0 +1,49 @@
...
+    },
+    {
+      file: isProd ? pkg.browser.replace('.js', '.min.js') : pkg.browser,
+      format: 'umd',
+      name: 'ReduxObservable',
+    },
+  ],
+  plugins: [
+    babel({
+      babelrc: false,
```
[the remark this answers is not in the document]
Shyam-Chen: `.babelrc` is used for testing. Use `babel-jest` can share `.babelrc`.

package.json
Shyam-Chen: Change `lib` to `dist`?

evertbouw:

Issue #446, opened 2018-03-09
Update TypeScript typings for more common use cases


ajcrites:
**Do you want to request a *feature* or report a *bug*?**  
I think this could be considered either, but I don't know the original philosophy of the typings.

This can possibly be broken out into multiple issues, but I think they are all related.

### Epic output action extends trigger action

`interface Epic<T extends Action, S, D = any, O extends T = T>` where `O extends T`. This means that any action emitted by the Epic has to conform to the structure of the action that triggers the epic. I think that it is very common to emit actions that have different properties. For example:

```ts
interface LoadUserAction {
  type: 'LOAD_USER';
  username: string;
}

interface LoadUserCompleteAction {
  type: 'LOAD_COMPLETE_USER';
  user: User;
}

const loadUserEpic: Epic<LoadUserAction, never, never, LoadUserCompleteAction>
```

This is not allowed because `'username'` doesn't exist on `LoadUserCompleteAction`.

### Update Epic generic type order for more common uses
It's common to have an epic that doesn't rely on the store at all, but if you want to add the generic type for trigger actions you have to set a type for the store state because it doesn't have a default.

I also don't think that options are commonly used, and in fact they're not even documented or properly typed (defaulting to `any`).

This leads me to use a lot of the following:

```ts
const epic: Epic<InputActions  | OutputActions, never, never, InputActions | OutputActions>
```

### `combineEpics` doesn't work well with epics using different types
The `combineEpics` generic typings only make sense if all of the Epics are of the same type, but this is uncommon -- at least compared to combining epics of different types.

```ts
const epic1: Epic<Epic1Trigger | Epic1Output, never>
const epic2: Epic<Epic2Trigger | Epic2Output, never>

const epics = combineEpics(epic1, epic2);
```

This is not allowed because the epic1 and epic2 actions do not conform to each others' types. To get around this you can circumvent type safety:

```ts
combineEpics<any>
combineEpics(...[epic1, epic2]);
```

or you have to be verbose

```ts
combineEpics<Epic1Trigger | Epic1Output | Epic2Trigger | Epic2Output, never>
combineEpics<typeof epic1 | typeof epic2>
```

## **What is the expected behavior?**
I have some suggested solutions, and I think that the proper solution should be to consider the most beneficial combination of ergonomics, type safety, and code completion.

* Epic output action:
Have `O` extend Action. It can default to `T` which keeps the same behavior as the current typings if it's not provided.

```ts
[...]

PR #473, opened 2018-04-10
feat(typings): Updating typings for more common use-cases.

ajcrites:
This is a first draft / suggestion for a possible reordering of typings re: #446 that accomplishes the following goals:

* Generic types for Epics are reordered to InputActions, OutputActions, State, Dependencies
* OutputActions now extend from Action instead of T allowing for more specific declarations for input and output actions
* State is made optional and defaults to 'void' for cases where it would be unused. This requires an explicit type for the generic if it's to be used
* Added a type for combineEpics that takes no generic arguments. It is common to have epics with different type signatures, and it would be very difficult / impossible to enforce type safety through generics for combineEpics at this level.

Tests have also been updated to use the new generic type ordering.

I think more changes are required for the tests with some other upcoming changes / support for Redux 4. I did the best that I could for now.

Re: @jayphelps's comments on #446, I understand what you mean about how all output actions are technically input actions as well ... but if I'm not mistaken _all_ actions / _any_ action that the store dispatches will be an input action filtered by `ofType`. My update is more for practical purposes as I see it:

* Ensure that input actions can only be of a certain type. Also provide code completion for input action properties.
* All epics must emit actions, so we have the ability to provide output actions to allow for code completion and consistency.
* State is less commonly used (input / output actions are always used). Default to `void` to enforce providing a valid state type when it is needed.
* Dependencies are least commonly used.

I think that these can be improved further, but this is what I've come up with so far based on my use cases. I definitely favor arranging the types in a way that are going to support community needs the most.


cjol:
From my use, I disagree with most of the motivation here. 

I'll start with what I do agree with: 
* Dependencies are the least commonly used, so I would support moving them to the back. 
* I'm happy to make State optional, though I prefer to do that by building my own type alias with my root state type baked in (see below). 
* It is beneficial from a type-safety perspective to specify an OutputAction type

Now my disagreements: 
* InputActions should indeed be _all_ actions that the store dispatches - anything else runs the risk of forgetting that `ofType` is required, which could add run-time errors that are not reflected by the static types.
* OutputActions are the only Type variable that needs stating (except dependencies) - InputActions and State are likely global across your application, so they should be baked into a type alias: 
`export type MyEpic<O extends AllActions = AllActions> = BaseEpic<AllActions, RootState, {}, O>;`

I think it might be helpful to demonstrate how I structure my Epics, so that you can see how I am able to get nice type safety and code completion. Note that I no longer use `ofType` because it doesn't seem to reliably narrow types - I could certainly look into building a PR that improves this behaviour of `ofType`.

```typescript
const loginRequest = () => ({ type: "LOGIN_REQUEST" as "LOGIN_REQUEST" });
const loginSuccess = (payload: TokenData) => ({
  type: "LOGIN_SUCCESS" as "LOGIN_SUCCESS",
  payload
});
const loginFailure = (payload: string) => ({
  type: "LOGIN_FAILURE" as "LOGIN_FAILURE",
  payload
});

type LoginRequest = ReturnType<typeof loginRequest>;
type LoginSuccess = ReturnType<typeof loginSuccess>;
type LoginFailure = ReturnType<typeof loginFailure>;

type AllActions = LoginRequest | LoginSuccess | LoginFailure;

type MyEpic<O extends AllActions> = Epic<AllActions, RootState, {}, O>;

const myEpic: MyEpic<LoginFailure | LoginSuccess> = (action$, state$) =>
  action$.pipe(
    filter((action): action is LoginRequest => action.type === "LOGIN_REQUEST"),
    map(x => loginFailure("Not implemented yet"))
  );
```

As a further note, in reality I actually reduce the boilerplate a little further by using the [`typesafe-actions`](https://github.com/piotrwitek/typesafe-actions) library. For example, the filter can be rewritten as simply `filter(isActionOf(loginRequest))` using `isActionOf` from the library.

Inside the `map`, the type of x has correctly been narrowed to LoginRequest. If I were to return something other than `LoginFailure | LoginSuccess`, myEpic would fail to type, so I feel confident in the safety of my code.

cjol:
If O extends T, then the pull request is just to re-order the type dependencies, right? 

In which case I think this PR is probably fine, as we don't lose any precision in our type definitions.

re: `ofType`, it doesn't seem to automatically narrow types for me. The next thing I tried (`isActionOf`) did work without any effort, so I never really spent the effort to see if I could make `ofType` work in the same way. I've noticed [this PR](https://github.com/redux-observable/redux-observable/pull/459) which might address it, but I haven't looked into it yet.

jayphelps:
Loving this discussion! @ajcrites did you have any follow up to @cjol's last comments?

ajcrites:
@jayphelps yes, perhaps I should have commented. Essentially I just implemented a reordering of the generic types to prioritize the more commonly used ones and provide sane defaults for the others.

test/typings.ts
```diff
@@ -22,23 +22,23 @@ interface Dependencies {
...
 }
 
-const epic1: Epic<FluxStandardAction, State> = (action$, store) =>
+const epic1: Epic<FluxStandardAction, FluxStandardAction, State> = (action$, store$) =>
   action$.pipe(
     ofType('FIRST'),
     mapTo({
       type: 'first',
-      payload: store.getState()
+      payload: store$.getState()
```
evertbouw: We probably shouldn't test for getState anymore. Can we change this to something like

```ts
const epic1: Epic<FluxStandardAction, FluxStandardAction, State> = (action$, state$) =>
  action$.pipe(
    ofType("FIRST"),
    withLatestFrom(state$),
    map(([, state]) => ({
      type: "first",
      payload: state
    }))
  );
```

ajcrites: @evertbouw since `getState` is not removed should we keep it? I wouldn't be opposed to adding a test for the `state$` stream too, though. Would you want that as part of this PR? It's not really related to the changes.

evertbouw: Yes fair enough. It should change soon but it doesn't have to be part of this PR

ajcrites:
@jayphelps I think we've agreed that the reorder is the way to go. From my end this is ready.

PR #486, opened 2018-04-27
feat(epicMiddleware): add infinite loop detection

### LINKED ISSUE #485: Feature Request: Add propper warning on missing emission of final action
When an action is unhandled (sometimes just one case is forgotten to be handled) or ignoreElements is not used the infinite loops created are not always easy to catch and prevents further actions to be handled.

Epected behaviour: 
throw error similar to: "Epic error: infinite loop detected because the action emitted the same action that it subscibed to."

Alternative: never emit the action that was subscribed to automatically somehow. Possibly add a hidden field before emitting the action to the action obeserver that would prevent handing of the same action twice. (if this behaviour is really wanted the same action could just be reemitted as a fresh action.)

evertbouw:
Added detection mechanism for infinite loops created in epics.

closes #485

@lucidNTR

### DIFF
--- a/src/createEpicMiddleware.js
+++ b/src/createEpicMiddleware.js
@@ -54,7 +54,22 @@ export function createEpicMiddleware(rootEpic, options = defaultOptions) {
         switchMap(output$ => options.adapter.output(output$))
       );
 
-      result$.subscribe(store.dispatch);
+      if (process.env.NODE_ENV !== 'production') {
+        // Check if actions returned from epics contain a property
+        // that only the actions going to epics should have.
+        // If present this indicates an infinite loop
+        result$.pipe(
+          require('rxjs/internal/operators/filter').filter(action => {
+            if (action.hasOwnProperty('@@ReduxObservable🦆') && action.type === action['@@ReduxObservable🦆']) {
+              require('./utils/console').warn(`Infinite loop detected. Action with type "${action.type}" was returned by an epic while it was also part of its input.`);
+              return false;
+            }
+            return true;
+          })
+        ).subscribe(store.dispatch);
+      } else {
+        result$.subscribe(store.dispatch);
+      }
 
       // Setup initial root epic. It's done this way so that
       // it's possible for them to call replaceEpic later
@@ -69,6 +84,16 @@ export function createEpicMiddleware(rootEpic, options = defaultOptions) {
         // It's important to update the state$ before we emit
         // the action because otherwise it would be stale!
         stateInput$.next(store.getState());
+
+        if (process.env.NODE_ENV !== 'production') {
+          // Add a hidden property to actions going to epics.
+          // This property should not be present in actions returned
+          Object.defineProperty(action, '@@ReduxObservable🦆', {
+            enumerable: false,
+            value: action.type,
+          });
+        }
+
         input$.next(action);
 
         return result;
--- a/test/createEpicMiddleware-spec.js
+++ b/test/createEpicMiddleware-spec.js
@@ -421,4 +421,38 @@ describe('createEpicMiddleware', () => {
     createStore(reducer, applyMiddleware(middleware));
     expect(epic.called).to.equal(true);
   });
+
+  it('should warn about inifite loops', () => {
+    sinon.spy(console, 'warn');
+
+    const epic = action$ => action$;
+    const reducer = state => state;
+    const middleware = createEpicMiddleware(epic);
+    const store = createStore(reducer, applyMiddleware(middleware));
+
+    store.dispatch({ type: 'ping' });
+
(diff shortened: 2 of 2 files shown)


test/createEpicMiddleware-spec.js
```diff
@@ -421,4 +421,38 @@ describe('createEpicMiddleware', () => {
...
+
+    console.warn.restore();
+  });
+
+  it('should not warn for no infinite loop', () => {
+    sinon.spy(console, 'warn');
+
+    const epic = action$ => action$.pipe(
+      ofType('ping'),
+      map(action => Object.assign(action, { type: 'pong' }))
```
evertbouw: How dirty do we need to get here. The hidden property should not be copied when using spread or an empty object as first arg

jayphelps:
I wonder if just checking for the same action being emitted once is a bit too aggressive? Someone may be doing this intentionally and it doesn't necessarily lead to it being infinite..though I haven't seen in practice.

I also wonder if us assigning a property to the action is going to bite someone some how, even if it isn't enumerable. They're still visible in the dev tools, which if anything is annoying IMO. We could keep a reference to the last action emitted, but this runs the issue of potentially being a notable memory leak. e.g. if the last action dispatched contains a giant payload, until the next action is received that memory won't be reclaimed. I doubt in practice it would be an issue, but the potential is something to consider.

I thought about keeping a reference to the _type_ of the last action emitted and a counter. If the same action is emitted N times synchronously in a row we throw. This is somewhat brittle since theoretically epics could be doing this intentionally, but again in practice probably aren't.

***

I think my biggest hangup is the assigning of the `'@@ReduxObservable🦆'` property to the action. This would annoy me if I saw it on my actions in the dev tools. 😆

src/createEpicMiddleware.js
```diff
@@ -54,7 +54,22 @@ export function createEpicMiddleware(rootEpic, options = defaultOptions) {
         switchMap(output$ => options.adapter.output(output$))
       );
 
-      result$.subscribe(store.dispatch);
+      if (process.env.NODE_ENV !== 'production') {
+        // Check if actions returned from epics contain a property
+        // that only the actions going to epics should have.
+        // If present this indicates an infinite loop
+        result$.pipe(
+          require('rxjs/internal/operators/filter').filter(action => {
```
jayphelps: We shouldn't reference operators via the internal path as it's subject to change without notice. I imagine you did it this way to avoid pulling it in on production builds, but I think it's unavoidable

jayphelps:
> The prop does not show up in the devTools for me.

http://jsbin.com/zofuqayazu/edit?js,output
image

***

> Now that I've had a while to think about it I'm not sure if this is the right approach.

I've had the same feeling thinking about this more, but at the same time if the same action is _synchronously_ emitted from the root epic I can't really imagine any real-world use case for it and even if there were, it's rare enough to me that as long as someone can opt-out of the warning warning about this is useful enough to first-timers that it's probably worth it.

I'm thinking perhaps what we should do is keep a reference to the action object itself we're going to provide to the root epic, if the root epic _synchronously_ emits the exact same action object we warn, then after the [`input$.next(action);`](https://github.com/redux-observable/redux-observable/blob/816a9163074943508de1300d3e1ca39603568186/src/createEpicMiddleware.js#L72) returns we clear the local reference to the action so that we don't leak it's memory and so we don't do that check for actions that come later async.

I do wonder if there are re-entrancy issues that could come about though--synchronous dispatches mean the middleware and then root epic will get re-entered before the original dispatch returns. It's also unclear whether we should actually prevent the action from being dispatched again or not. If we don't, we'd have to do something clever so that we don't keep warning for every single action until the callstack size is blown cause that would be thousands of warnings.

rgbkrk:
If you use a `Symbol` and use a `get`ter the property won't show up in the devtools. It's also only accessible through `Object.getOwnPropertySymbols(action)`:

screen shot 2018-05-18 at 9 38 06 am

Probably don't want to do this still.

deanius:
As desirable as this seems to be for some use cases, I think it’s against the spirit of the library to disallow, or add loop-policing time costs toward ‘loop detection’. Think about the analog on a programming language: a ‘while(true)’ loop or a function calling itself both may produce infinite loops- but are valid tools. I don’t think the infinite loop is something to be policed at runtime as much as caught in code review. On a trampolining or other kind of schedule, there’s not even a risk of blowing the stack. Just my two cents.

evertbouw:
withdrawn

deanius:
I learned some stuff in the discussion though, and I am watching the march toward 1.0 very closely :)

jayphelps:
1.0 has been released actually 😁

deanius:
Ok, not that closely I guess @jayphelps ☺️ Well, cool 🎉 !!

PR #493, opened 2018-05-15
feat(createEpicMiddleware): schedule emitted actions and epic subscription on the queueScheduler

jayphelps:
BREAKING CHANGE:

You must now provide your rootEpic to `epicMiddleware.run(rootEpic)` instead of passing it to `createEpicMiddleware`. This fixes issues with redux v4 where it's no longer allowed to dispatch actions while middleware is still being setup. Redux v4 support will be coming in a follow up PR.

BREAKING CHANGE:

## Actions emitted by your epics are now scheduled on a queue

In 1.0.0 we now subscribe to your root Epic, and dispatch actions emitted by it, using the queueScheduler from RxJS. This is a bit hard to explain (and understand) but as the name suggests, a queue is used. If the queue is empty, the action is emitted as usual, but if that action synchronously causes other actions to be emitted they will be queued up until the call stack of the first action returns.

In a large majority of cases this will have no perceivable impact, but it may affect the order of any complex epic-to-epic communication you have.

The benefit is that actions which are emitted by an epic on start up are not missed by epics which come after it. e.g. With `combineEpics(epic1, epic2)` previously if epic1 emitted on startup, epic2 would not receive that action because it had not yet been set up.

Take this example:

```ts
const epic1 = action$ =>
  action$.pipe(
    ofType('FIRST'),
    mergeMap(() =>
      of({ type: 'SECOND' }, { type: 'THIRD' })
    )
  );

const epic2 = action$ =>
  action$.pipe(
    ofType('SECOND'),
    map(() => ({ type: 'FOURTH' })),
    startWith({ type: 'FIRST' })
  );

// notice that epic2 comes *after* epic1
const rootEpic = combineEpics(epic1, epic2);
```

In older version of redux-observable, your reducers would have been missing the FOURTH:

```
FIRST
SECOND
THIRD
```

However in 1.0.0 it now would see it as the last one:
```
FIRST
SECOND
THIRD
FOURTH
```

In that example, the SECOND action is now seen by epic2 because it is queued on the same schedule as subscribing (setting up) the Epics themselves is. Since the middleware will try to subscribe to the Epics first, it now always will finish doing so before any action is emitted--so epic2 doesn't miss any actions.

Another way of looking at it is that when an individual Epic is synchronously emitting actions, they will always be emitted in the sequence provided, without any other Epics being able to sneak another action in-between. When we did `of({ type: 'SECOND' }, { type: 'THIRD' })`, we now know _for sure_ that THIRD will immediately follow SECOND; in older versions of redux-observable this wasn't guaranteed as another Epic could have been listening for SECOND and emitted some other action before THIRD, because they shared the same call-stack.

Because this is dealing with very complex recursion, call stacks, and sequences, this may be tough to fully wrap your head around. We hope that what actually happens in practice is itself more intuitive, even if truly understanding how things are queued is now.

I'm *very* open if someone has a better way of explaining this. FWIW redux-saga and ngrx both do queue scheduling as well, though the behavior is different between us three in certain edge cases.


jayphelps:
I'm a bit conflicted about the queue scheduling behavior. I personally think people will find it more intuitive in their average code, but it is true that if you have a _very_ good understanding and mental picture of call stacks, this behavior would in fact end up being very unintuitive. While most of us would hope we have that level of understanding, in practice I've seen not only myself but many others struggle with these epic-to-epic interactions.

I'm curious what others think. Should we keep things as-is, go this route, or something else?

jayphelps:
Here's a jsbin that lets you play around and compare the behavior of the new/old:

https://jsbin.com/buvibaf/1/edit

Change `useQueueScheduling = false` in the HTML tab to switch. Play around with various complex and recursive epic-to-epic communication.

jayphelps:
## synchronous emit on startup

Currently the order of your epics provided to combineEpics matters not just by the order in which they'll receive actions (unavoidable as single threaded) but it also means that if an epic emits synchronously on start up, any epic that comes after it won't receive it because they are not yet subscribed to.

```ts
const epic1 = action$ =>
  action$.pipe(
    ofType('SOMETHING'),
    startWith({ type: 'FIRST' })
  );

const epic2 = action$ =>
  action$.pipe(
    ofType('FIRST'),
    map(() => ({ type: 'SECOND' }))
  );

const rootEpic = combineEpics(epic1, epic2);
```

#### Current

```
FIRST
```

#### With queue scheduling

```
FIRST
SECOND
```

```ts
const epic1 = action$ =>
  action$.pipe(
    ofType('FIRST'),
    mergeMap(() =>
      of({ type: 'SECOND' }, { type: 'THIRD' })
    )
  );

const epic2 = action$ =>
  action$.pipe(
    ofType('SECOND'),
    map(() => ({ type: 'FOURTH' })),
    startWith({ type: 'FIRST' })
  );
```

jayphelps:
# Examples

Here are some examples to help grok the differences in complex epic-to-epic communication. On one hand, the current behavior is technically "correct" my feeling is that it isn't as intuitive as the behavior would be with the queueScheduler.

I'm very curious if others agree/disagree.

## synchronous emit on startup

Currently the order of your epics provided to combineEpics matters not just by the order in which they'll receive actions (unavoidable as single threaded) but it also means that if an epic emits synchronously on start up, any epic that comes after it won't receive it because they are not yet subscribed to.

```ts
const epic1 = action$ =>
  action$.pipe(
    ofType('SOMETHING'),
    startWith({ type: 'FIRST' })
  );

const epic2 = action$ =>
  action$.pipe(
    ofType('FIRST'),
    map(() => ({ type: 'SECOND' }))
  );

const rootEpic = combineEpics(epic1, epic2);
```

#### Current

```
FIRST
```
(no SECOND action because epic2 isn't listening in time to catch FIRST)

#### With queue scheduling

```
FIRST
SECOND
```

## synchronously emitting sequential actions

In the current version if an epic tries to emit more than one action synchronously and sequentially, it's possible that another epic could be listening and emit it's own action before the originating epic has finished dispatching it's own.

```ts
const epic1 = action$ =>
  action$.pipe(
    ofType('FIRST'),
    mergeMap(() =>
      of({ type: 'SECOND' }, { type: 'THIRD' })
    )
  );

const epic2 = action$ =>
  action$.pipe(
    ofType('SECOND'),
    map(() => ({ type: 'FOURTH' }))
  );
```

#### Current

```
FIRST
SECOND
FOURTH
THIRD
```

(FOURTH is before THIRD)

#### With queue scheduling

```
FIRST
SECOND
THIRD
FOURTH
```

This behavior could also been seen in even more complicated scenarios where an epic is listening for the action it also emits itself:

```ts
const epic1 = action$ =>
  action$.pipe(
    ofType('FIRST'),
    mergeMap(() =>
      merge(
        action$.pipe(
          ofType('SECOND'),
          map(() => ({ type: 'FIFTH' }))
        ),
        of({ type: 'SECOND' }, { type: 'THIRD' })
      )
    )
  );

const epic2 = action$ =>
  action$.pipe(
    ofType('SECOND'),
    map(() => ({ type: 'FOURTH' }))
  );
```

#### Current

```
FIRST
SECOND
FOURTH
FIFTH
THIRD
```

(FOURTH and FIFTH is before THIRD)

#### With queue scheduling

```
FIRST
SECOND
THIRD
FOURTH
FIFTH
```

## common takeUntil patterns may also not subscribe early enough

Common patterns I've seen using `takeUntil` also currently have the issue where they are unintuitively not yet subscribed when someone thinks they are.

```ts
const epic1 = action$ =>
  action$.pipe(
    ofType('FIRST'),
    mergeMap(() =>
      timer(100).pipe(
        map(() => ({ type: 'FOURTH' })),
        takeUntil(action$.pipe(
          ofType('THIRD')
        )),
        startWith({ type: 'SECOND' })
      )
    )
  );

const epic2 = action$ =>
  action$.pipe(
    ofType('SECOND'),
    map(() => ({ type: 'THIRD' }))
  );
```

#### Current

```
FIRST
SECOND
THIRD
FOURTH
```

(FOURTH is emitted because the takeUntil wasn't yet subscribing to see that THIRD was dispatched by epic2)

#### With queue scheduling

```
FIRST
SECOND
THIRD
```

(FOURTH is not emitted because that Observable was cancelled because epic2 emitted THIRD)

jayphelps:
> Or is it just because you are comfortable working with a stack and a queue requires a different mental model?

I definitely still get bitten by deep epic-to-epic callstack complexities, but I've run into it enough times helping others that I often can spot the pattern quickly. My thought was that this queue usage would make things more intuitive for the average user however there's a decent risk that this will actually trip someone up who is expecting the previous behavior. I'm also afraid that this is a higher risk because I haven't figured out a great way to explain it without just showing numerous before/after examples, which makes me second guess this.

I think I'll keep this open for a few days or a week for comments and if there are no objections go ahead and merge and release it as another alpha to see if we broke people.

jayphelps:
Thanks to everyone for discussions and reviews.

@Andarist thanks again for discussing with me in DMs and giving insight into what's worked for redux-saga. <3

squgeim:
I am sorry to revive a closed issue -- we just upgraded redux-observable in our project and this change has broken some of our features. I was hoping there would have been discussions on work-arounds for situations that rely on this, but I haven't found any.

The scenario in our system is that we have an epic that dispatches a list of actions, each of those actions have an epic of their own that return actions that ultimately updates the store. Some of the epics in that list expect that an action before it has "completed executing" (ie, their epic has executed and made changes to the store).

After the upgrade, the situation is that all the actions in the list are dispatched in a single batch, their respective epics are called in order, but the actions returned by these epics have not been dispatched. This breaks the assumption that an epic in the list can expect that an epic before it has completed making changes to the store.

What do you guys think about this use case? Is this usage a bad practice? How do you suggest I reimplement this? Or is there a way to make sure I force it to "commit" an epic before executing the next one.

jayphelps:
@squgeim can you make a new ticket with any minimal reproduction(s)? Feel free to use your preferred environment, but Stack Blitz makes it pretty easy: https://stackblitz.com/fork/redux-observable-playground

squgeim:
> @squgeim can you make a new ticket with any minimal reproduction(s)? Feel free to use your preferred environment, but Stack Blitz makes it pretty easy: https://stackblitz.com/fork/redux-observable-playground

I have asked it as a question on Stack Overflow.
https://stackoverflow.com/questions/62584821/how-to-make-sure-one-epic-is-completed-before-executing-the-next-epic

PR #501, opened 2018-05-24
chore(redux): Finalize Redux 4 support

evertbouw:
Bumped version number and fixed tests

### DIFF
--- a/test/initAction.d.ts
+++ b/test/initAction.d.ts
@@ -1,0 +1,1 @@
+export const initAction: { type: string };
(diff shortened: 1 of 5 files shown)


test/initAction.d.ts
```diff
@@ -0,0 +1 @@
+export const initAction: { type: string };
```
Ailrun: Is there a way to change this `string` to more specific one?
I'm afraid that this would mess up some union type like
```
// Following will be just a string
type actionType = (myAction1 | myAction2 | initAction)['type'];
```

Ailrun: Oh, sorry. It's just for the test. My mistake.

tcozzens:
This would be great!

jayphelps:
I changed the commit to be a feat() and added a BREAKING CHANGE block, both so that it shows up in the generated CHANGELOG

PR #503, opened 2018-05-31
fix(state$): initial state is no longer skipped, and remove deprecated store APIs

jayphelps:
I found that state$ was actually skipping the initial first state and in fact because of the way redux v4 works and our new `epicMiddleware.run(rootEpic)` API

BREAKING CHANGE: state$ now receives the initial state

BREAKING CHANGE: previously the second argument to your epics was a "lite" version of the redux store with store.dispatch() and store.getState(), however this has been replaced with a stream of a state$ in v1.0.0 of redux-observable and the old methods were deprecated with warnings in previous alpha versions. This release removes them entirely. See https://redux-observable.js.org/MIGRATION.html

### DIFF
--- a/test/createEpicMiddleware-spec.js
+++ b/test/createEpicMiddleware-spec.js
@@ -92,6 +92,10 @@ describe('createEpicMiddleware', () => {
 
     expect(store.getState()).to.equal(2);
     expect(actions).to.deep.equal([initAction, {
+      type: 'PONG',
+      input: 0,
+      state: 0
+    }, {
       type: 'PING'
     }, {
       type: 'PONG',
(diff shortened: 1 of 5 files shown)


jayphelps:
Okiedoke I went through some revisions on this but I think it's ready at least to get it out to unblock those who run into this bug. So I believe this is ready for review and merge if good.

***

There's a bigger follow up discussion about whether or not this means we can now use a BehaviorSubject instead of a custom StateObservable. We _can_ but the only thing stopping us now is that it seems like a footgun to expose it to epics because while we want `state$.value` we _don't want_ `state$.next()` error, complete, unsubscribe that are also on BehaviorSubject.

test/createEpicMiddleware-spec.js
```diff
@@ -190,57 +175,6 @@ describe('createEpicMiddleware', () => {
     }]);
   });
 
-  it('should warn about accessing state$.value before @@redux/INIT', () => {
```
evertbouw: Is there a way to keep this? It works now and there is probably a proper use case for it but it's going to cause confusion. We should add it to the FAQ at least

jayphelps: ohhh like keep the test but invert it? That you _can_ access `state$.value` on epic startup because INIT has already been passed through reducers?

jayphelps: I went ahead and added that. lmk if you actually wanted something else

jayphelps: oh, I don't want to warn because this is now perfectly valid. It's true that often it might be a mistake, but I'm not sure we can both offer it as valid _and_ warn?

jayphelps: Gonna merge as we can introduce the warning again if we figure out a good way. Definitely interested in continuing this discussion

evertbouw: Similar situation as the infinite loop then. Will need to add it to the FAQ

jayphelps:
Released as 1.0.0-beta.0

PR #512, opened 2018-06-04
fix(combineEpics): combineEpics no longer errors on React Native Android because of readonly `name`  prop setting

jayphelps:
Reported by someone on Discord, relayed by @evertbouw. On React Native Android it appears the JS runtime enforces the fact that Functions are supposed to have their `name` property be readonly, by throwing an exception. So in those cases we just catch the error and ignore it since I'm not aware of any other way to test whether it's supported or not. I'm also not sure how we could unit test this behavior either.

image

### DIFF
--- a/src/combineEpics.js
+++ b/src/combineEpics.js
@@ -14,7 +14,14 @@ export const combineEpics = (...epics) => {
     })
   );
 
-  return Object.defineProperty(merger, 'name', {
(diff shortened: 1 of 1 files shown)


evertbouw:
LGTM

jayphelps:
FWIW I could not reproduce this in the latest version of Safari, which uses JavaScriptCore, which is what React Native uses on Android. However, the version it uses is likely much much older and also contains other changes to it so I can't be certain.

PR #518, opened 2018-06-06
fix(errors): errors from reducers are no longer caught and logged, instead are rethrown

jayphelps:
This PR is against the `stable` branch (tracking 0.18.0), not `master` (tracking 1.0.0-beta.1). It is not applicable to master, so it does not need to be ported.

***

related #263#issuecomment-395109222

BREAKING CHANGE: For 0.19.0 errors from reducers are no longer caught and console.error logged, instead they are just rethrown as before. This was a temporary workaround for a bug in rxjs where it would silently swallow errors. That bug has been fixed in 5.5.6+, so it is highly recommended you use _at least_ rxjs@5.5.6+ with this version of redux-observable. However, redux-observable is close to reaching 1.0.0-final which will require rxjs v6 and redux v4, if you'd like to start upgrading to it now you can use redux-observable@next (as of this writing 1.0.0-beta.1)

I thought about bumping the peer dep to rxjs@5.5.6 but figured it might be overkill since it still works with previous versions, it just runs the risk of potentially swallowing errors. Does anyone think we _should_ bump the peer dep to restrict it?

Cc/ @anton164

### DIFF
--- a/src/createEpicMiddleware.js
+++ b/src/createEpicMiddleware.js
@@ -53,13 +53,7 @@ export function createEpicMiddleware(rootEpic, options = defaultOptions) {
           return output$;
         })
         ::switchMap(output$ => options.adapter.output(output$))
-        .subscribe(action => {
-          try {
-            store.dispatch(action);
-          } catch (err) {
-            console.error(err);
-          }
-        });
+        .subscribe(store.dispatch);
 
       // Setup initial root epic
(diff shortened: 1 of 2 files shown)


jayphelps:
ignore the commit-lint failure. Was trying it out, but I disagree with the restriction on commit message length.

PR #531, opened 2018-06-15
feat(adapters): remove adapters support

jayphelps:
BREAKING CHANGE: Support for adapters has been removed. Adapters were previously used to transform the `action$` Observable into some other stream-library primitive; like Most.js, Bacon, RxJS v4, etc. While rarely used, if you would like this functionality the MIGRATION.md guide gives an example

***

This feature was added early when RxJS v4 was still dominant. Based on npm install counts of the public adapters it seemed to be used by a very very small number of people. I'm certainly open to hearing arguments for keeping it, in which case we'll need to figure out how the adapter is applied to the new `state$` Observable.

You can achieve the effect of adapters by applying the transforms in a custom root Epic. Here's an example of converting the Observables to Most.js streams:

```js
import most from 'most';
import { from } from 'rxjs';

// a Most.js implementatin of combineEpics
const combineEpics = (...epics) => (...args) =>
  most.merge(
    ...epics.map(epic => epic(...args))
  );

const rootEpic = (action$, state$, ...rest) => {
  const epic = combineEpics(epic1, epic2, ...etc);
  // action$ and state$ are converted from Observables to Most.js streams
  const output = epic(
    most.from(action$),
    most.from(state$),
    ...rest
  );

  // convert Most.js stream back to Observable
  return from(output);
};
```


docs/FAQ.md
```diff
@@ -1,34 +0,0 @@
...
-
-##### most.js
-##### bacon.js
-##### etc.
-
-It should be possible for you to create an adapter to add support. See [redux-observable-adapter-rxjs-v4](https://github.com/redux-observable/redux-observable-adapter-rxjs-v4) for reference.
-
-<a id="miscellaneous-thunkservables-deprecated"></a>
-<a id="why-were-thunkservables-removed"></a>
-### Why were thunkservables removed?
```
jayphelps: I removed this too as it's old enough to not matter I think

Issue #502, opened 2018-05-25
[PROPOSAL] redux-saga like server side rendering


hitmands:
**Do you want to request a *feature* or report a *bug*?**  
Feature

**What is the current behavior?**
A running epic cannot be terminated due to `epicMiddleware#run` missing its complementary API `epicMiddleware#terminate`

**What is the expected behavior?**
As a user I expect to run/terminate any epic.

**Which versions of redux-observable?**
`latest`

**Potential Usages**
1. **Server Side Rendering** where you need to perform some operations before serving the html to the browser
2. **Code Splitting** where you need to add/remove epics on-the-fly (redux provides the [`replaceReducer`](https://redux.js.org/recipes/code-splitting#using-replacereducer) API).

In https://github.com/redux-observable/redux-observable/pull/539 I am proposing an actual implementation:

```typescript
interface EpicMiddleware {
  // sends termination signal via the actions stream (can be caught and ignored),
  kill(Epic): Observable,

  // kills the process and cannot be caught.
  sigterm(Epic): Observable,
}
```

### Resources

- https://www.google.com/search?q=difference+between+sigterm+and+sigkill&oq=difference+between+sigter&aqs=chrome.0.0j69i57j0l6.4509j1j7

evertbouw:
This proposal is quite different from https://github.com/redux-observable/redux-observable/pull/412, now I'm not using SSR myself so I don't know what's required for it, but I'd love to get a discussion going!

ping @Jerry-Hong @chiel

hitmands:
Hello, 
I was just wondering if this might be falling into an oblivion...

please, let me know how can I help.
Giuseppe

jayphelps:
I think @BerkeleyTrue has experience with using redux-observable in SSR. Maybe he has thoughts?

deanius:
I think I follow - you want a way to hook the end of all epics triggered by initializing an app, because only then will all its data be ready? And your application of that is for SSR. 

If that's true, I think stating this issue in terms of that might draw out a little more discussion..

I've been thinking about this type of issue for some time - I've been working on solving it outside of RO.. But basically I think you're more or less on the same train as me: 

If you have a list of epics you're waiting on, then the `forkJoin` of them contains the results of all their completions. So when that observable emits, that's your hook. The key is to know which renderers are set into flight by each action (the initialization action included). And for that you may manually track that in a read-only, non-serializable property of the `action.meta`, or in some other place.

[...]

PR #539, opened 2018-06-24
WIP: provide server side rendering support via sigterm/sigkill

hitmands:
### Ticket
- https://github.com/redux-observable/redux-observable/issues/502
___

___

### Outstanding
- [ ] typing
- [ ] handle not-found threads

### DIFF
--- a/src/createEpicMiddleware.ts
+++ b/src/createEpicMiddleware.ts
@@ -1,12 +1,21 @@
 import { Action, Middleware, MiddlewareAPI, Dispatch } from 'redux';
-import { Subject, from, queueScheduler } from 'rxjs';
-import { map, mergeMap, observeOn, subscribeOn } from 'rxjs/operators';
+import { Subject, from, queueScheduler, Observable } from 'rxjs';
+import { finalize, map, mergeMap, observeOn, subscribeOn, takeUntil } from 'rxjs/operators';
 import { StateObservable } from './StateObservable';
 import { Epic } from './epic';
 import { warn } from './utils/console';
 
 interface Options<D = any> {
   dependencies?: D;
+}
+
+interface ThreadStorage {
+  [key: string]: null | {
+    id: string;
+    sigterm$: Subject<void>;
+    sigkill$: Subject<void>;
+    output$: Observable<any>;
+  };
 }
 
 export interface EpicMiddleware<
@@ -16,6 +25,8 @@ export interface EpicMiddleware<
   D = any
 > extends Middleware<{}, S, Dispatch<any>> {
   run(rootEpic: Epic<T, O, S, D>): void;
+  sigterm(rootEpic: Epic<T, O, S, D>): Observable<O>;
+  sigkill(rootEpic: Epic<T, O, S, D>): Observable<O>;
 }
 
 export function createEpicMiddleware<
@@ -41,6 +52,7 @@ export function createEpicMiddleware<
 
   const epic$ = new Subject<Epic<T, O, S, D>>();
   let store: MiddlewareAPI<Dispatch<any>, S>;
+  const threads: ThreadStorage = {};
 
   const epicMiddleware: EpicMiddleware<T, O, S, D> = _store => {
     if (process.env.NODE_ENV !== 'production' && store) {
@@ -62,23 +74,40 @@ export function createEpicMiddleware<
 
     const result$ = epic$.pipe(
       map(epic => {
-        const output$ = epic(action$, state$, options.dependencies!);
+        const id = epic.id;
+        const sigterm$ = new Subject<void>();
+        const sigkill$ = new Subject<void>();
 
-        if (!output$) {
+        const thread$ = epic(
+          action$.pipe(
+            takeUntil(sigterm$),
+          ),
+          state$,
+          options.dependencies!,
+        );
+
+        if (!thread$) {
           throw new TypeError(
             `Your root Epic "${epic.name ||
               '<anonymous>'}" does not return a stream. Double check you\'re not missing a return statement!`
           );
         }
 
-        return output$;
+        return { id, thread$, sigkill$, sigterm$ };
       }),
-      mergeMap(output$ =>
-        from(output$).pipe(
+      mergeMap(({ id, thread$, sigkill$, sigterm$ }) => {
+        const output$ = from(thread$).pipe(
           subscribeOn(uniqueQueueScheduler),
-          observeOn(uniqueQueueScheduler)
-        )
-      )
+          observeOn(uniqueQueueScheduler),
+          takeUntil(sigkill$),
+          finalize(() => {
+            threads[id] = null;
+          }),
+        );
+
+        threads[id] = { id, output$, sigkill$, sigterm$ };
+        return output$;
+      })
     );
 
     result$.subscribe(store.dispatch);
@@ -106,7 +135,48 @@ export function createEpicMiddleware<
         'epicMiddleware.run(rootEpic) called before the middleware has been setup by redux. Provide the epicMiddleware instance to createStore() first.'
       );
     }
+
+    if (!rootEpic.id) {
+      Object.defineProperty(rootEpic, 'id', {
+        value: Math.random().toString(16),
+      });
+    }
+
     epic$.next(rootEpic);
+  };
+
+  epicMiddleware.sigterm = (rootEpic) => {
+    const id = rootEpic.id;
+    const thread = threads[id];
+
+    if (!thread) {
+      // thread is not running,
+      // probably need to throw
+      return;
+    }
+
+    setTimeout(() => {
+      thread.sigterm$.next();
+    }, 0);
+
+    return thread.output$;
+  };
+
+  epicMiddleware.sigkill = (rootEpic) => {
+    const id = rootEpic.id;
+    const thread = threads[id];
+
+    if (!thread) {
+      // thread is not running,
+      // probably need to throw
+      return;
+    }
+
+    setTimeout(() => {
+      thread.sigkill$.next();
+    }, 0);
+
+    return thread.output$;
   };
 
   return epicMiddleware;
--- a/test/createEpicMiddleware-spec.ts
+++ b/test/createEpicMiddleware-spec.ts
@@ -2,7 +2,7 @@ import 'babel-polyfill';
 import { expect } from 'chai';
 import sinon from 'sinon';
 import { createStore, applyMiddleware, Reducer, Middleware, Action, AnyAction } from 'redux';
-import { createEpicMiddleware, combineEpics, StateObservable, ofType, Epic, __FOR_TESTING__resetDeprecationsSeen as resetDeprecationsSeen } from '../';
+import { createEpicMiddleware, combineEpics, StateObservable, ofType, Epic, __FOR_TESTING__resetDeprecationsSeen as resetDeprecationsSeen } from '../src';
 import { of, empty, merge, queueScheduler, Observable } from 'rxjs';
 import { mapTo, filter, map, mergeMap, startWith, ignoreElements, distinctUntilChanged } from 'rxjs/operators';
 import { initAction } from './initAction';
@@ -447,6 +447,93 @@ describe('createEpicMiddleware', () => {
       ]);
 
       done();
+    });
+  });
+});
+
+describe('createEpicMiddleware#{sigkill,sigterm}', () => {
+  let context: any;
+
+  beforeEach(() => {
+    const reducer: Reducer = (state = [], action) => state.concat(action);
+    const epic1: Epic = action$ =>
+      action$.pipe(
+        ofType('INPUT'),
+        mapTo({ type: 'E1-OUTPUT' }),
+      );
+    const epic2: Epic = action$ =>
+      action$.pipe(
+        ofType('INPUT'),
+        mapTo({ type: 'E2-OUTPUT' }),
+      );
+
+    const middleware = createEpicMiddleware();
+    const store = createStore(reducer, applyMiddleware(middleware));
+
+    const verifyActions = (given: { type: string }[]) => {
+      expect(store.getState()).to.deep.equal([initAction, ...given]);
+    };
+
+    context = { middleware, store, epic1, epic2, verifyActions };
+  });
+
+  it('should complete after sigterm', (done) => {
+    const { epic1, epic2, middleware, store, verifyActions } = context;
+
+    queueScheduler.schedule(() => {
+      middleware.run(epic1);
+      middleware.run(epic2);
+
+      expect(epic1).to.have.property('id');
+      expect(epic2).to.have.property('id');
+
+      store.dispatch({ type: 'INPUT' });
+
+      const thread$ = middleware.sigterm(epic1);
+
+      thread$.toPromise().then(() => {
+        store.dispatch({ type: 'INPUT' });
+
+        verifyActions([
+          { type: 'INPUT' },
+          { type: 'E1-OUTPUT' },
+          { type: 'E2-OUTPUT' },
+          { type: 'INPUT' },
+          { type: 'E2-OUTPUT' },
+        ]);
+
+        done();
+      }).catch(done);
+    });
+  });
+
+  it('should complete after being killed', (done) => {
+    const { epic1, epic2, middleware, store, verifyActions } = context;
+
+    queueScheduler.schedule(() => {
+      middleware.run(epic1);
+      middleware.run(epic2);
+
+      expect(epic1).to.have.property('id');
+      expect(epic2).to.have.property('id');
+
+      store.dispatch({ type: 'INPUT' });
+
+      const thread$ = middleware.sigkill(epic1);
+
+      thread$.toPromise().then(() => {
+        store.dispatch({ type: 'INPUT' });
+
+        verifyActions([
+          { type: 'INPUT' },
+          { type: 'E1-OUTPUT' },
+          { type: 'E2-OUTPUT' },
+          { type: 'INPUT' },
+          { type: 'E2-OUTPUT' },
+        ]);
+
+        done();
+      }).catch(done);
     });
   });
 });


hitmands:
https://github.com/redux-observable/redux-observable/issues/502

index.d.ts
```diff
@@ -39,6 +39,7 @@ export declare interface Epic<Input extends Action = any, Output extends Input =
 
 export interface EpicMiddleware<T extends Action, O extends T = T, S = void, D = any> extends Middleware {
   run(rootEpic: Epic<T, O, S, D>): void;
+  terminate(): Promise<undefined>;
```
Brooooooklyn: `Promise<void>`

src/createEpicMiddleware.js
```diff
@@ -72,5 +80,18 @@ export function createEpicMiddleware(options = {}) {
     epic$.next(rootEpic);
   };
 
+  epicMiddleware.terminate = () => {
+    if (process.env.NODE_ENV !== 'production' && !store) {
+      require('./utils/console').warn('epicMiddleware.terminate() called before the middleware has been setup by redux. Provide the epicMiddleware instance to createStore() first.');
+    }
+
+    if (!terminating) {
+      terminating = new Promise((resolve) => { resolveTermination = resolve; });
```
Brooooooklyn: I think using `Subject` is more elegant here
And type signature in `EpicMiddleware` would be `terminate(): Observable<void>`, it's more consistent for redux-observable

index.d.ts
hitmands: [] use-void

test/createEpicMiddleware-spec.js
```diff
@@ -414,4 +418,38 @@ describe('createEpicMiddleware', () => {
...
+      );
+
+    const middleware = createEpicMiddleware();
+    const store = createStore(reducer, applyMiddleware(middleware));
+
+    middleware.run(combineEpics(
+      createFakeEpic('baz', 20),
+      createFakeEpic('bar', 10),
+      createFakeEpic('foo'),
+    ));
```
hitmands: this test would only pass as long as all the epics will return an `Observable` from `action$`...

adding something like this would cause it to fail:

```typescript
// this would not be terminated by action$
    const epicFromState = (action$, state$) => state$
      .pipe(
        filter(state => state.length),
        delay(100),
        map(payload => ({ type: 'log', payload }))
      );
```

src/createEpicMiddleware.js
hitmands: - [ ] use-subject

src/createEpicMiddleware.js
```diff
@@ -72,5 +80,22 @@ export function createEpicMiddleware(options = {}) {
     epic$.next(rootEpic);
   };
 
+  epicMiddleware.terminate = () => {
```
hitmands: not currently supporting multiple calls to `middleware.run`

src/createEpicMiddleware.js
```diff
@@ -72,5 +80,22 @@ export function createEpicMiddleware(options = {}) {
...
+      require('./utils/console').warn('epicMiddleware.terminate() called before the middleware has been setup by redux. Provide the epicMiddleware instance to createStore() first.');
+    }
+
+    if (!terminating) {
+      terminating = new Subject();
+      resolveTermination = () => {
+        terminating.complete();
+      };
+
+      actionSubject$.complete();
```
hitmands: we probably have to terminate `stateSubject$` as well.

test/createEpicMiddleware-spec.js
hitmands: may `stateSubject.complete()` handle this case?

src/createEpicMiddleware.js
hitmands: - [ ] support-multiple-iterations

hitmands:
This pr implements a method `terminate` which is in opposition to `run`, so every call to `epicMiddleware.terminate()` will cause the current call to `epicMiddleware.run()` to terminate. 

expected:
```javascript
// epics will start retrieving actions
epicMiddleware.run(rootEpic);

// close actions stream and wait for all epics to terminate their current iteration
await epicMiddleware.terminate();

// your state is now correctly populated;
ReactDOMServer.renderToNodeStream(<Provide store={store}><App /></Provider>);
```

cons: since it relies on `action$`, every returned `epic` should terminate when `action$` does it.
@Brooooooklyn any thought on how to avoid this dependency?

jayphelps:
Awesome! Thanks for implementing this so it's exactly clear how it would work.

I'm gonna think about this solution a bit more, but one immediate thing that comes to mind is that it seems a bit odd that you can run() any number of epics concurrently but only terminate all of them at the same time instead of independently. Maybe this isn't a problem, maybe it is. It also seems odd that once you terminate you can't run any other epics, but I have no proof that in practice it would be useful.

I generally like the idea of the action$ stream given to an epic complete()ing as the signal epics use to "shut down" as it allows them to potentially finish up any work they need to do, including async work, instead of being abruptly killed. In this case I'm using process/daemon terms since epics are conceptually very similar.

It may be useful to take process the analogy further in that OS processes usually have signals you can send to a process to either terminate gracefully (e.g. SIGTERM) or kill it immediately without grace (e.g. SIGKILL). Maybe there's merit in adopting these as redux actions? It might be overcomplicating things, though it seems sorta neat that an epic could terminate easily. I've often wondered if we should expose a way for epics to spawn other epics a la carte. I don't wanna get ahead of myself though and overcomplicate your SSR use case, just an anecdote.

hitmands:
Hi Jay, 

> It also seems odd that once you terminate you can't run any other epics, but I have no proof that in practice it would be useful.

I think this is sorted now, there is a [spec here](https://github.com/redux-observable/redux-observable/pull/539/files#diff-435ff52c600611ea18f12c1c3f399ea9R452).

The concept of iteration is intended per `run`, so, 
you could potentially invoke as many `run` as you want.

```typescript
epicMiddleware.run(rootEpic1);
await epicMiddleware.terminate().toPromise();

epicMiddleware.run(rootEpic2);
await epicMiddleware.terminate().toPromise();
```

hitmands:
> I generally like the idea of the action$ stream given to an epic complete()ing as the signal epics use to "shut down" as it allows them to potentially finish up any work they need to do, including async work, instead of being abruptly killed.

Yeah, that was my initial purpose.
The `action$ completion` will serve as a `SIGTERM` for your epic, so, any epic knows when to terminate.

Should I change my current implementation in favor of an action like `@@redux-fluent/iteration/sigterm` ?

hitmands:
> it seems a bit odd that you can run() any number of epics concurrently but only terminate all of them at the same time instead of independently.

actually we can only send a `SIGTERM` (read `action$ completion`) but then it is up to your epic to terminate, we will only wait for that [`finalize`](https://github.com/redux-observable/redux-observable/pull/539/files#diff-3d5499bd89cd728e032282fe927ce951R48)

Brooooooklyn:
hi @hitmands 
> since it relies on action$, every returned epic should terminate when action$ does it.

As I was going deeper into your question, I realized there's one situations that not covered 

considering there is a `epic` that would trigger another epic:

```ts
const initEpic = action$ =>
  action$.pipe(
    ofType(INIT_ACTION)
    exhaustMap(() => /** something async */) 
    map(INIT_SUCCESS)
    endWith(GET_INIT_DATA)  // terminate here, and dispatched GET_INIT_DATA would trigger nothing
  )

const getInitDataEpic = (action$, state$) =>
  action$.pipe(
    ofType(GET_INIT_DATA)
    withLatestFrom(state$)
    switchMap(([_, state]) => /** something async */ )
    map(GET_INIT_DATA_SUCCESS)
  )

export const epic = combineEpics(initEpic, getInitDataEpic)

```

In this situation, we could not find a way to **guess** when should we `takeUntil` the `action$`

hitmands:
hi @Brooooooklyn,
I see your point. This case wouldn't be covered unless yo'd call `epicMiddleware.terminate()` at the end of your second epic (or eventually merge those two epics).

We could consider to leave this kind of logic to the developer...

I can't think of anything to handle this situation and having every epic to have to terminate itself could result a bit verbose and lead to the necessity of having dedicated epics between client and server...

hitmands:
Hi @jayphelps,
is there anything I could do to help this thread?

jayphelps:
@hitmands Sorry that this has sat unmerged. To be honest I haven't ever needed SSR for redux-observable so I was hoping for more opinions from the community that needs this.

> actually we can only send a SIGTERM (read action$ completion) but then it is up to your epic to terminate, we will only wait for that finalize

Yeah, that's what feel off to me. I would imagine that you could terminate individual epics, perhaps run() returns a subscription for that individual epic that you can call unsubscribe on to kill it?

***

Is there anyone subscribed to this thread that can chime in with their thoughts on the API this PR provides?

d1ab:
Is there any chance that this feature will be merged in next release? I'm probably not the only one can not wait for SSR to arrive.

vendramini:
I'm using `next.js` with `redux-observables`. There's a workaround for SSR, but native implementation would be nice.

d1ab:
> I'm using `next.js` with `redux-observables`. There's a workaround for SSR, but native implementation would be nice.

@vendramini can you privde more details about that walkaround?

vendramini:
> > I'm using `next.js` with `redux-observables`. There's a workaround for SSR, but native implementation would be nice.
> 
> @vendramini can you privde more details about that walkaround?

@d1ab sure.

There's an "oficial implementation" here: https://github.com/zeit/next.js/tree/canary/examples/with-redux-observable

But I don't like the idea to instantiate the `StateObservable` and call the `rootEpic` then dispatch the action. Imagine doing this in every page. Not cool.

All this because we MUST return a `Promise` from dispatch(), so `getInitialProps` could `await` for the response.

My solution: `redux-thunk` with `redux-observable`.

I created this function:
```javascript
const promised = action => dispatch => new Promise((resolve, reject) => dispatch({...action, resolve, reject}));
export default promised;
```
Then I wrap my action with it, transforming the action in a thunk:
```javascript
export const fetchAreas = () => promised(actions.areas.fetch());
```
Notice that in `promised` I pass the `resolve` and `reject`, so I can handle it inside my epic:
```javascript
export const fetchAreasEpic = (action$: ActionsObservable) => action$.pipe(
    ofType(FETCH),
    mergeMap(
        action => httpGet('areas').pipe(
            map(ajaxResponse => fetchAreasSuccess(ajaxResponse.response)),
            tap(next => next.error ? action.reject() : action.resolve())
        )
    ),
);
```
What annoys me in my solution: in every epic that is promised, you must call: `tap(next => next.error ? action.reject() : action.resolve())` to handle the Promise.

But now you could do this on SSR:
```javascript
if (!store.getState().productsHighlights)
    {
        await Promise.all(
            [
                store.dispatch(fetchProductsHighlights()),
                store.dispatch(fetchServicesHighlights()),
                store.dispatch(fetchCompaniesPopular()),
                store.dispatch(fetchPostsHighlights()),
            ]
        );
    }
```
Not the ideal world, but much better.

What do you think? Any issue ask me.

jayphelps:
The primary thing we need before merging any SSR features is consensus for a number of people in the community around what behavior it should have. Once it's added, it's very hard to take it away or change it in a breaking way, so we default to not providing something since it is possible to still do it independently.

Some comments and questions regarding this particular PR were made above, asking for community feedback if anyone wants to chime in on this particular one. Other proposals are also welcome, but we'd need community consensus and collecting all the folks can be a bit tough--maybe talking to folks on the Gitter channel would help.

fuusle:
I have used redux-observable in a React SSR application (via the react-redux-epic library). I have found the limitation/simplicity of not being able to do multiple "levels" of dispatches (epic-to-epic) to be adequate and ultimately, a good thing. Adequate because the server-side stuff typically tends to be more static (content-wise) and simple (for example, user-bound logic and data are off-limit on the server-side (in my experience)). Also, you probably want the server-side-render to be as fast as possible, anyway. Forcing developers to keep things simple and preventing chains of actions to be dispatched is arguably a good thing.

I like the approach in this PR. It's simple and has a small conceptual footprint 😄

I would suggest two things, though:

**1) Reconsider renaming the `terminate()`** method to something like `.shutdown()` or `.end()`. If I understand the implementation correctly, the inner observables of the epics are allowed to finish already started initiatives(?)  If so, `terminate()` feels "too strong" (like `kill`/`pkill`), but `shutDown`/`end` implies a kind of graceful shutdown procedure. 

**2) The terminate()-method should always return a Promise?**
It looks like the interaction of the terminate-method is going to be in the context of the server request-handler, more or less exclusively. Using promise directly, simply feels cleaner.  Also, I consider the `iterationSubject$` to be internal to the epic-middleware. 

Further, I think the terminate()-method should never return null (as it will when no actions are dispatched).  It should just return a resolved promise (`Promise.resolve`) instead.

fuusle:
@jayphelps, would it help to think of the functionality of this PR to not (necessarily) be dedicated to SSR? I mean, the only thing this PR provides, is a way to gracefully stop listening to new actions after the `.terminiate()` is called. This is more than enough to get started with basic SSR (for the reason stated in my previous comment). 

Further, you already have `middleware.run(myRootEpic)`, so it kind of makes sense that you can `middleware.terminate().then(() => {})' 

Even further, if the community would like the middleware to dispatch some kind of "last call" action as discussed in the comments above. That could be considered a "phase two" that could be implemented without breaking the "contract" given in this PR.

jayphelps:
@fuusle I'm cool with examining it that way. I've been thinking on this and think what we might want to do is have `middleware.run(epic)` return a Subscription, and you can call `sub.unsubscribe()` on it to shut it down. That way you have control over each time you do `middlewware.run(epic)` and don't necessarily _have_ to shut them all down at the same time. If you could then compose subscriptions, if you indeed wanted to shut them down at the same time.

```js
const composite = new Subscription();

const sub1 = middleware.run(epic1);
composite.add(sub1);

const sub2 = middleware.run(epic2);
composite.add(sub2);

composite.unsubscribe();
```

Anyone have thoughts?

jayphelps:
Unsubscribing from the run() wouldn't do anything except what you'd expect calling unsubscribe() on the result of the epic's return observable subscription. So it would **NOT** dispatch any actions, or anything like that. If you wanted to signal to your epics that a "shutdown" is about to happen, you could dispatch your own action easily. This is inline with Redux Observable's general philosophy that it is mostly meant as a way to reduce boilerplate to integrate with Redux and RxJS, not intended to be an opinionated framework. This is similar to how you could recreate the gist of Redux with RxJS trivially, but you miss out on its niceties.

It would be nearly the same conceptually (without RO's niceties) as this:

```js
const action$ = new Subject();
const sub = rootEpic(action$).subscribe(action => store.dispatch(action));
store.subscribe(action => action$.next(action);

// later
sub.unsubscribe();
```

aplassen:
@jayphelps, thinking in terms of unsubscribing, would be equally nice. 

What would happen to epics that are "working" on fetching something, would they be allowed to dispatch the resulting action (typically a success or fail action) to the redux store?

```
             
          fetchingResources
          _________
         /         \
---A-----B-----|----C------
               ^
               unsubscribe/terminate point
```

**C** needs to be dispatched in the Redux store (but need not reach other epics)

aplassen:
Indeed, I appreciate the efforts to keep redux-observable unopinionated. In fact, that was one of the reasons we choose the library (get reusable reactive/rxjs skills) 👍

jayphelps:
@aplassen 

> What would happen to epics that are "working" on fetching something, would they be allowed to dispatch the resulting action (typically a success or fail action) to the redux store?

They would not, at the point of unsubscribe(). It's important to think of this as "just RxJS" so same as if you unsubscribe from any other stream, it will propagate that unsubscribe to the whole chain, and if any of the operators have teardown logic (like fromFetch/ajax do) then they will be torndown/cancelled.

Not to worry though, you can send a signal that shut down is about to happen, and do something in response to that if you'd like. However, I'm not sure when this would be idiomatic. Do you have any examples? In the example you provided "would they be allowed to dispatch the resulting action (typically a success or fail action) to the redux store" this would not be idiomatic because you're shutting down, you shouldn't care anymore about the fetch request and so you _should_ want it to be cancelled. If that's not the case, I'm misunderstanding the use case?

Also--since this is regular RxJS for the most part, you can also do things inside a `finalize(() => stuff)` operator, but it does not allow you to next() anything because it's already too late to do so. It's only for imperative side effect clean up, if you happened to do some of those (avoid when possible)

Here's an example, where I implemented a fake createEpicMiddleware that has this unsubscribe behavior as a prototype:

https://stackblitz.com/edit/redux-observable-playground?file=index.js

aplassen:
Good point! Unsubscribing would, of course, mean exactly that. Which would not support the use case for SSR (see the provided example under). 

Thanks for clarifying where you are coming from with a stackblitz!

I'll do my best not to conflate the idiomatic goal of redux-observable with the implementation details of SSR. To illustrate the mechanism that would be needed to use redux-observable in SSR, I have compiled a  small overview:
 
## Concrete use case for SSR (React-based)
On the server you have to render in a synchronous way, making it hard to hydrate the application with data because of the asynchronous nature of fetching data. Several strategies exists: 

- Predict the route at hand (e.g. /harddrives/ssd/232) and do the fetching before you render the React application, ensuring that the data can be access in a synchronous manner.  
- Others like **Apollo**, parses the React-tree, extracting and hydrating  the data-queries before the application is rendered (getDataFromTree()) 
- **Next.js** uses a getInitialProps()-method on so-called "Pages" (a concept in Next.js) that the server part of Next.js knows how to handle

**Finally, the one method used in react-redux-epic,** is the one I have experience with. The idea is simple: 

- Render the application in two phases
   - The first render will trigger a set of Redux dispatches, requesting data
   - In the second phase, said requests for data have hydrated the Redux store, and the route can be rendered
- The hydrated store is also part of the request (ready to hydrate the store on the client)
- You have close to no difference out in the application in regards to client or server
   - ...but server-side epics that gets invoked are simple as possible (someAction -> fetch -> someSuccessAction)

_(Note: it's worth mentioning that this is behind a cache, so the extra cost of two renders are quite acceptable)_

Quick code to illustrate how the original PR could be used instead of react-redux-epic:

```js
// First render: dispatch any requests for data/state
ReactDOMServer.renderToStaticMarkup(component);

epicMiddleware
  .termiate() // takeUntil is triggered, preventing any more actions to be
  .toPromise()
  .then(() => {
    // Second render: now with a hydrated Redux store
      const markup = ReactDOMServer.renderToString(component)
      // ... (send markup to clent)
  });
```

aplassen:
I guess the "idiomatic blueprint" for the pattern could be illustrated with something like:

```js
const lastCall = (
  of(1)
    .pipe(
      delay(2000),
      tap(() => console.log('lastCall fired'))
    )
)

interval(1000)
  .pipe(
    takeUntil(lastCall),
    mergeMap(() => (
      of('FETCH')
        .pipe(
          delay(5000), // the "fetch" are not finished when the lastCall notifications come
          mapTo({ type: 'FETCH_SUCCEEDED' }) // this is allowed to be a notification
        )
    )),
    tap((action) => {
      console.log(action)
    })
  )
  .subscribe(() => {}, () => {}, () => { console.log('Done') })
// lastCall
// { type: 'FETCH_SUCCEEDED' }
// Done
```

jayphelps:
@aplassen Gotcha, you want the original strategy of complete()ing the action$ and letting the root epic some how signal "ok I'm done". We'd have to complete() state$ too probably. It's also possible for some epic to never terminate, so I'd highly advise adding a timeout.

SSR seems like the only reason you'd want to do this AFAICT and it's also pretty easy to implement yourself, which is why I've been inclined not to provide it built-in, unless we can get consensus from several people in the community. The reason is because once we ship something, it's pretty difficult to change it because people will rely on it. Because I don't personally do SSR with RO I haven't been able to experiment in practice to confirm this approach works, so I have to rely on the community.

https://stackblitz.com/edit/redux-observable-playground-skmxy3?file=index.js

hitmands:
What about extending the `epicMiddleware` interface by adding:

```
interface WithSSR {
  // dispatches a sigterm action so every epic 
  // can finish their current iteration and then complete
  sigterm(): void

  // alias of unsubscribe, 
  // kills all the epics straight away
  kill(): void;
}
```

1. use case will be used to guarantee one iteration during server side rendering
2. would be called after some timeout...

aplassen:
@jayphelps 

> It's also possible for some epic to never terminate, so I'd highly advise adding a timeout. 
Indeed! 👍

> https://stackblitz.com/edit/redux-observable-playground-skmxy3?file=index.js
Wow! That's a pretty neat solution 👏

I suspect the only trouble would be the`.value` prop to access the current state value since we are returning a plain `Observable` (and not a `StateObservable`)?

```
 // ...
      state$.pipe(takeUntil(shutdown$)), // returns a plain observable, right?
/// ...
```

aplassen:
@hitmands

Could you hit me with some examples or use cases on the server where you have epics that you want to terminate / notify about cancelation? 

Did you have a look at 
https://stackblitz.com/edit/redux-observable-playground-skmxy3?file=index.js 
what are your thoughts on this approach? 

Btw, thanks for this PR, it's a great initiative! 👏

jayphelps:
>  suspect the only trouble would be the.value prop to access the current state value since we are returning a plain Observable (and not a StateObservable)?

Yep you're right. I don't think StateObservable is designed in a way to complete() right now ever. There's ways around this, but as far as what I'd recommend I'd have to play around and I don't have cycles at the moment unfortunately. Open to suggestions if someone has! e.g. you could do this, but I dunno if there are caveats:

```js
    import { StateObservable } from 'redux-observable';
    
    const newState$ = new StateObservable(
      state$.pipe(takeUntil(shutdown$)),
      state$.value
    );
```

liamzdenek:
I have used redux-observable for a small/unimportant SSR application before, but this was about two years ago. I've used redux-observable quite a bit in the meantime, however. I do have some thoughts about this that I hope aren't too late to the conversation (apologies, I meant to finish this comment last weekend, but life got in the way)

**1) Observables which do not propagate `action$` completion to the top-level/returned epic [will fail to trigger this finalize operator](https://github.com/redux-observable/redux-observable/pull/539/commits/4ac50d635446978a7d4069f60f1f83b7b36aae0a#diff-3d5499bd89cd728e032282fe927ce951R48)**, which causes `iterationSubject$` to never `.complete()`, deadlocking any subscribers to the `.terminate()` call (and presumably causing the SSR to time out). Some example epics which should exhibit this behavior:

```js
// this is an admittedly antagonistic example
function e1(action$, state$) { return Observable.NEVER; }

// if you replaced ignoreElements with map/filter calls, the .terminate() behavior should be unchanged
function e2(action$, state$) { return state$.pipe( ignoreElements() ) }
```

Adding similar behavior to `state$` as was added for `action$` will solve part of this, but not all of it. Users can put whatever observable logic they wish inside of an epic. Users can write a valid observable that does not use either `state$` or `action$`. I think a minimum non-deadlocking solution would involve the existing solution, plus proper support for closing `state$`, plus adding a timeout to the rootEpic, using a time limit provided eg: `.terminate(10000)`. I'm not sure of something more elegant but I definitely want something more elegant.

**2) Users may wish to call `.terminate()` multiple times** if their requests have not been dispatched by the time the first `.terminate()` call completes. (I'm not saying such a move is necessarily wise, just that a use case exists here) Take the following hypothetical epics, which are a mirror of a common pattern in many redux-observable applications:

```js
const authenticationEpic = (action$, state$) => action$.ofType(DO_AUTH).pipe(
  switchMap(_unused =>
    // performs an API call, and emits AUTH_COMPLETE
  )
);

const profileEpic = (action$, state$) =>action$.ofType(AUTH_COMPLETE).pipe(
  switchMap(_unused =>
    // this part doesn't matter for this example
  )
);
```

If `.terminate()` is called before AUTH_COMPLETE is dispatched, then the profileEpic will not be executed. Users may wish to call `.terminate()` repeatedly until their desired set of data is present. But, I don't really like this implementation for that purpose.

If I step a bit back and think about what I really want here -- it's that I want to have some arbitrary logic that monitors the action stream and state store, and independently determines when to call terminate. I think this could be neatly modeled as an epic that is passed to the terminate call:

```js
middleware.terminate((action$, state$) => {
  return action$.ofType(AUTH_COMPLETE).pipe(take(1));
});
```

I haven't totally thought through every edge case here, but here's roughly what I'm thinking: The terminate function body would subscribe to the provided observable. When the observable emits a value, that triggers `iterationSubject$.next();` in the same way the current implementation works.

Providing an interface this rich is probably overkill for most users, so shipping a well-thought-out default terminate epic is probably worth considering if this approach is used.

---

More broadly, I would love a library that does this, but I heavily lean in favor of keeping core libraries such as redux-observable as thin as possible until there's a specific need. I think all of this can be implemented in user space using the currently implemented interfaces, so I'd prefer to play this safe and keep this unmerged for the time being.

jayphelps:
I generally agree with those points. As far as this living in a community library, that might be better for quicker experimentation, but I think the biggest thing we need to figure out the suggested way for people to have StateObservable complete() since it is not designed to do so currently. A library could implement their own StateObservable (it's not complex) if they wanted to move fast today and be unblocked, but ultimately RO should modify its own in some way. Probably what it should do is propagate error/complete from the input subject it takes in. Easy enough seemingly.

aplassen:
@jayphelps 

> There's ways around this, but as far as what I'd recommend I'd have to play around and I don't have cycles at the moment unfortunately. 

No worries! This thread has proven to be quite insightful and I think, at least for my simple use case, the examples you provided, prove that this can be solved in application space.

It's worth mentioning that the use case I presented is more inline with the Next.js approach to "getInitialProps()". More "long-lived" epics, exists and runs on the client. In fact, only epics related to the "getInitialProps()-approach" need to run on the server. "Universal Epics" as it were 😉
 
## Though experiment 
In the application space one could also implement some kind of factory/wrapper that handles the timeouts and other safe guards / distingue you would have for your epics running on the server:

```js
const getInitialProductsEpic = universalEpic((action$) => (
  action$
    .pipe(
      ofType('PRODUCTS_REQUESTED'),
      mergeMap((category) => (
        fetchProducts(category)
          .pipe(
            map((products) => ({ type: 'PRODUCTS_FETCH_SUCCEEDED', products })),
            catchError((error) => of({ type: 'PRODUCTS_FETCH_FAILED', error }))
          )
      ))
    )
));
```

aplassen:
@liamzdenek 

Thanks for leaving your thoughts in this thread 👏  

Regardless of the feasibility, I though it was a cool way of looking at it (returning an action to signal when to actually terminate).

jaybekster:
Good day!
Are the any updates and how far is it from being finished?

aplassen:
@jaybekster, I suspect that this PR is not going to be be merged. 

However, depending on you needs, you can use redux-observable with SSR. Look at [this comment above](https://github.com/redux-observable/redux-observable/pull/539#issuecomment-543186192) with a [stackblitz exmaple](https://stackblitz.com/edit/redux-observable-playground-skmxy3?file=configureStore.js).

hitmands:
@aplassen I have updated the pr to show how could we implement a sigterm/kill strategy for the given epics. Please take a look

cc: @jayphelps

src/createEpicMiddleware.ts
```diff
@@ -109,5 +125,19 @@ export function createEpicMiddleware<
...
+  };
+
+  epicMiddleware.kill = rootEpic => {
+    // this should only kill the given epic, alias of `complete`.
+  };
+
+  epicMiddleware.killall = () => {
+    // running epics could be stored into a list when calling `middleware.run`
+    // this method could flush that list and call `complete`
+  };
```
hitmands: `sigterm` could just work inside the redux life cycle, so all currently running epics could complete their current iteration.

`kill` and `killall` would just terminate everything

src/createEpicMiddleware.ts
```diff
@@ -106,8 +122,29 @@ export function createEpicMiddleware<
...
```
hitmands: this would address give the ability of killing/terminating one or all the currently running epics

hitmands: Implementation details:

1. assign an `ID` to every epic passed to `middleware.run`
2. store every epic passed to `middleware.run` into a long-living variable
3. use the ID to terminate a specific epic
4. use the list of running epic to terminate everything indiscriminately

src/createEpicMiddleware.ts
```diff
@@ -106,8 +132,37 @@ export function createEpicMiddleware<
...
+
+    if (!thread) { /* epic has never been run */ }
+
+    setTimeout(() => {
+      thread.kill$.next();
+    }, 0);
+
+    return thread.output$;
+  };
+
```
hitmands: ### Preconditions

1. every running epic (thread) should be assigned a thread `id` to allow the middleware (thread manager) to terminate/kill threads individually

### Interface

1. **sigterm**: terminates the actions stream so that the thread would still have the time to complete any on-going job (e.g. async tasks), emit their last output and then exit gracefully
2. **kill**: terminates the thread directly, any on-going job would be abruptly killed straight away.

### Potential Usage

```typescript
const index = async (req, res) => {
  epicMiddleware.run(rootEpic);
  store.dispatch({ type: 'FETCH_TODOS' });
  
  const thread$ = epicMiddleware.sigterm(rootEpic);
  await thread$.toPromise();


  ReactDOMServer.renderToNodeStream(<Provider store={store}><App /></Provider>).pipe(res);
};
```

### Pros

1. Being able to manage every process individually (`run`, `kill`, `terminate`).

### Cons

1. need for extra `killall` and `termall` methods to kill and terminate all the currently running threads
2. if a rootEpic is run twice, then `kill` and `terminate` would affect all their instances (not necessarily a problem though)
3. `sigterm` can only terminate the action stream, so epics not propagating this completion would end up in a deadlock (as brilliantly highlighted by @liamzdenek [here](https://github.com/redux-observable/redux-observable/pull/539#issuecomment-543479927))

### Workarounds

1. Implement timeout to kill deadlocking processes

```typescript
const timeout = (middleware, rootEpic) => new Promise((resolve) => {
  setTimeout(() => {
    const thread$ = middleware.kill(rootEpic);
    thread$.toPromise().then(resolve);
  }, 2000);
});

const index = async (req, res) => {
  epicMiddleware.run(rootEpic);
  store.dispatch({ type: 'FETCH_TODOS' });
  
  const thread$ = epicMiddleware.sigterm(rootEpic);
  await Promise.race([
    thread$.toPromise(),
    timeout(epicMiddleware, rootEpic),
  ]);


  ReactDOMServer.renderToNodeStream(<Provider store={store}><App /></Provider>).pipe(res);
};
```

___

### Motivations

While SSR is probably one of the main use cases for implementing such API, it would also find applications in projects that heavily rely in code splitting. 

I am currently working on a project that mounts/unmounts reducers, components and epics depending on routing.

While we can easily add/remove reducers and components at runtime, it is impossible to do so with epics, since they cannot be terminated once ran.

**It really feels like `run` is missing its complementary api `terminate`**. (e.g. [redux#replaceReducer](https://redux.js.org/api/store#replacereducernextreducer))

src/createEpicMiddleware.ts
```diff
@@ -106,8 +138,49 @@ export function createEpicMiddleware<
...
   };
 
+  epicMiddleware.sigterm = (rootEpic) => {
+    const id = rootEpic.id;
+    const thread = threads[id];
+
+    if (!thread) {
+      // thread is not running,
+      // probably need to throw
+      return;
```
hitmands: or any observable that never completes...

src/createEpicMiddleware.ts
```diff
@@ -106,8 +135,49 @@ export function createEpicMiddleware<
...
+    }
+
+    setTimeout(() => {
+      thread.sigkill$.next();
+    }, 0);
+
+    return thread.output$;
+  };
+
```
hitmands: ### Preconditions

1. every running epic (thread) should be assigned a thread `id` to allow the middleware (thread manager) to terminate/kill threads individually

### Interface

1. **sigterm**: terminates the actions stream so that the thread would still have the time to complete any on-going job (e.g. async tasks), emit their last output and then exit gracefully
2. **sigkill**: terminates the thread directly, any on-going job would be abruptly killed straight away.

### Potential Usage

```typescript
const index = async (req, res) => {
  epicMiddleware.run(rootEpic);
  store.dispatch({ type: 'FETCH_TODOS' });
  
  const thread$ = epicMiddleware.sigterm(rootEpic);
  await thread$.toPromise();


  ReactDOMServer.renderToNodeStream(<Provider store={store}><App /></Provider>).pipe(res);
};
```

### Pros

1. Being able to manage every process individually (`run`, `kill`, `terminate`).

### Cons

1. if a rootEpic is ran multiple times, then `sigkill` and `sigterm` would affect all their instances (not necessarily a problem though)
2. `sigterm` can only terminate the action stream, so epics not propagating this completion would end up in a deadlock (as brilliantly highlighted by @liamzdenek [here](https://github.com/redux-observable/redux-observable/pull/539#issuecomment-543479927))

### Workarounds

1. Implement timeout to kill deadlocking processes

```typescript
const ensureShutdown = (middleware, rootEpic) => {
  setTimeout(() => {
    middleware.sigkill(rootEpic);
  }, 2000);
};

const index = async (req, res) => {
  epicMiddleware.run(rootEpic);
  store.dispatch({ type: 'FETCH_TODOS' });
  
  const thread$ = epicMiddleware.sigterm(rootEpic);
  ensureShutdown(epicMiddleware, rootEpic)
  await thread$.toPromise();


  ReactDOMServer.renderToNodeStream(<Provider store={store}><App /></Provider>).pipe(res);
};
```

### Additional (probably needed) steps

- `sigterm` should also send the signal to the `StateObservable`
- add methods to sigkill/sigterm all the running epics at once.

___

### Motivations

While SSR is probably one of the main use cases for implementing such API, it would also find applications in projects that heavily rely in code splitting. 

I am currently working on a project that mounts/unmounts reducers, components and epics depending on routing.

While we can easily add/remove reducers and components at runtime, it is impossible to do so with epics, since they cannot be terminated once ran.

**It really feels like `run` is missing its complementary api `terminate`**. (e.g. [redux#replaceReducer](https://redux.js.org/api/store#replacereducernextreducer))

### Should this feature live in RO or be implemented in a separate package?

It is my opinion that a complementary api for `run` is needed to reach feature completeness, nevertheless, shipping first-class support for SSR would drive adoption :)

oste:
> > suspect the only trouble would be the.value prop to access the current state value since we are returning a plain Observable (and not a StateObservable)?
> 
> Yep you're right. I don't think StateObservable is designed in a way to complete() right now ever. There's ways around this, but as far as what I'd recommend I'd have to play around and I don't have cycles at the moment unfortunately. Open to suggestions if someone has! Maybe BehaviorSubject as a workaround..

Has there been a workaround for this? I went so far as to implement [this](https://stackblitz.com/edit/redux-observable-playground-skmxy3?file=configureStore.js) solution as an example in the official next.js project the inability to access `$state.value` in epics seems like a limiting factor. 

Please see the diff I would like to introduce here https://github.com/vercel/next.js/compare/canary...oste:redux-observable-example-ssr#diff-4dea9d50fa13117bdce5cfe8ccaa457a5893b80949ad61cef1aa9259637635a5

The error is `TypeError: Cannot read property 'nextUserId' of undefined` and is the result of this code https://github.com/oste/next.js/blob/3513ae92175df97e836cb925ab132d3864dd9d94/examples/with-redux-observable/store/epics.js#L27

Unfortunately, my knowledge is still limited on how to best resolve this @jayphelps, @hitmands, @aplassen, @evertbouw, @Jerry-Hong 

Since Next.js seems to be taking a clear lead as the de facto SSR framework it would be good to provide a definitive example of how SSR can be handled there.

evertbouw:
I agree with that. I've just started a project with Nextjs so I might be able to contribute something. I have no need for redux (-observable) in there right now tho

oste:
@evertbouw That would be great. You should be able to pull my branch and run the redux-observable example. Just need to change to that directory and run `yarn install` and then `yarn next dev` and you should be up and running and able to reproduce the issue.

ttiurani:
> Has there been a workaround for this? I went so far as to implement this https://stackblitz.com/edit/redux-observable-playground-skmxy3?file=configureStore.js solution as an example in the official next.js project, however, the inability to access `$state.value` in epics seems to be a limiting factor.

I'm running into this same problem too with `$state.value` missing. Were you able to work around it somehow @oste ?

PR #559, opened 2018-08-07
feat(StateObservable): Add getState method to StateObservable

naderhen:
This would be extremely helpful when migrating to the latest versions of redux-observable. We are in a weird spot where our application relies on RxJS 6, but previously only supported redux-observable@0.x. I understand if this is philosophically against what you'd like for the library, but I thought it might be helpful for other just so that they can migrate incrementally. Maybe a 1.0.0-compat version? Please let me know if there's anything else I can do/provide in the meantime.

### DIFF
--- a/src/StateObservable.js
+++ b/src/StateObservable.js
@@ -23,4 +23,11 @@ export class StateObservable extends Observable {
       }
     });
   }
+
+  getState() {
+    if (process.env.NODE_ENV !== 'production') {
+      require('./utils/console').warn('Deprecated Usage of getState Found.');
+    }
+    return this.value;
+  }
 }


evertbouw:
If incremental upgrades is all you needed, you can upgrade to v1 alpha.2 before upgrading to latest. There the state observable still had this method

naderhen:
Thanks @evertbouw, I'm trying that locally with my existing codebase and I seem to get an error with the following epic structure:

```
const myEpic = (action$, {getState}) =>
    action$
        .filter(() => {
            return someSelector(getState());
        })
        .ofType(ACTION)
        .mergeMap(() => Rx.Observable.empty());
```

image


It doesn't seem like `this` is being properly set. If I don't use destructuring, and instead access `state$.getState()` directly it seems to work. Not sure if I missed a migration step for this version

evertbouw:
Hmmm yes I can see why that happens. When it was the (partial) Redux store getState was just a function but with state$ it is a method. Your proposed change should also suffer from this. Looks like you will have to replace all of those with state$.value to get it working again
________________________________

PR #575, opened 2018-09-26
docs(ErrorHandling.md): Fix Error Handling in RxJS

clitetailor:
The right way to handle error based on RxJS internal docs: https://github.com/ReactiveX/rxjs/blob/6.3.2/src/internal/operators/catchError.ts#L15-L69

You may need to update the example too!

### DIFF
--- a/docs/recipes/ErrorHandling.md
+++ b/docs/recipes/ErrorHandling.md
@@ -6,21 +6,26 @@ This can be done with the `catchError()` RxJS operator:
 
 ```js
 import { ajax } from 'rxjs/ajax';
+import { of, concat } from 'rxjs';
 
 const fetchUserEpic = action$ => action$.pipe(
   ofType(FETCH_USER),
-  mergeMap(action => ajax.getJSON(`/api/users/${action.payload}`).pipe(
-    map(response => fetchUserFulfilled(response)),
-    catchError(error => of({
-      type: FETCH_USER_REJECTED,
-      payload: error.xhr.response,
-      error: true
-    }))
-  ))
+  mergeMap(action => ajax.getJSON(`/api/users/${action.payload}`)),
+  map(response => fetchUserFulfilled(response)),
+  catchError((error, caught) =>
+    concat(
+      of({
+        type: FETCH_USER_REJECTED,
+        payload: error.xhr.response,
+        error: true
+      }),
+      caught
+    )
+  )
 );
 ```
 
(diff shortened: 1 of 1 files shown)


jayphelps:
Hey there. Thanks for the PR!

I actually don't think the catch should be moved outside the mergeMap because I believe it's less understandable but more importantly because it actually can potentially have different, unintended behavior because if you let the error escape the mergeMap and catch it top level the epic is effectively "terminated" in the sense that the error causes unsubscribes all the way up to action$. Then inside the catchError you provided `caught` (the source the operator is applied to) meaning that it will resubscribe effectively "restarting" the epic. Aside from wasteful teardown/setup of extra subscriptions, if your epics performs side effects on "start up" you'll perform them again, or if you maintain state you'll lose it.

```js
const fetchUserEpic = action$ => action$.pipe(
  ofType(FETCH_USER),
  throttleTime(300), // <---------- HERE
  mergeMap(action => ajax.getJSON(`/api/users/${action.payload}`)),
  map(response => fetchUserFulfilled(response)),
  catchError((error, caught) =>
    concat(
      of({
        type: FETCH_USER_REJECTED,
        payload: error.xhr.response,
        error: true
      }),
      caught
    )
  )
);
```

In this example, our throttleTime operator maintains state. That state will be thrown away every time there is an ajax error if we placed the catchError in that spot.

Sorry if this is confusing! It's tough to explain. Let me know if I also misunderstood your intent.

clitetailor:
Yeah, thanks for the tip. However, I think it's the rare case to have `throttleTime` come before `catchError`. I means i understand that it's common for the ajax to be failed but i think it is the thing that we have to work with the stream itself than to ignore it. For example, this is a work around:

```jsx
concat(
      of({
        type: FETCH_USER_REJECTED,
        payload: error.xhr.response,
        error: true
      }),
      delay(300),
      caught
    )
```

It may not fully reset the state of the stream but i think i will try to find the better solution for this. However it is good to let other ones know that this thing exists! 😄

About the cancellation example, i think there's a bleeding edge too. We not sending the fully cancellation request but instead ignore the result. Is it good? I'm i missing something. So if you know more i hope that you can explains!

clitetailor:
Humm, i think you are right about handing error. Should we add some caution about this to the document?

clitetailor:
Okey, i found how to deal with this:

```jsx
.pipe(
  mergeMap(x => of(1 / x).pipe(
    catchError(err => of(err))
  ))
)
```

This should be how we handling the side effects. Should i make some caution or posting a question to RxJS issues? 🤣

clitetailor:
However, it's still good to use `caught` we don't store any state stream at all. So usage might come with cautions!

PR #598, opened 2019-01-02
feat(createEpicMiddleware): inject the action & state in each dependency

eyal0803:

I have a use case where I need to conditionally add an HTTP header to each request based on a value in my Redux store. To solve this problem I discovered `dependencies` which I can pass to `createEpicMiddleware` and access in each epic. So I thought maybe I'd pass the following `dependencies`:
```js
{
    get: ajax.getJSON
}
```

But then I realized that I'd still have to manually (check & )pass the header to the function, so that didn't solve it.

My addition iterates over the passed `dependencies`, check whether each is a function, and if so it wraps it in a function that passes it the `state$` and the `action$` (which I don't need for my use-case but might as well pass it (or not?)).

If there's any other solution for my use case that doesn't require this change please let me know, it would be much appreciated.

**Edit:** Not sure why but the `npm run build` doesn't pass, even without the changes.

### DIFF
--- a/src/createEpicMiddleware.js
+++ b/src/createEpicMiddleware.js
@@ -28,6 +28,21 @@ export function createEpicMiddleware(options = {}) {
 
     const result$ = epic$.pipe(
       map(epic => {
+        // Interate over the dependencies and pass the action and the state to
+        // each function
+        for (const dependency in options.dependencies) {
+          if (options.dependencies.hasOwnProperty(dependency)) {
+            const dependencyValue = options.dependencies[dependency];
+
+            // Alter the dependency only if it's a function
+            if (typeof dependencyValue === 'function') {
+              // Pass the action and the state to the function
+              options.dependencies[dependency] = () =>
+                dependencyValue(action$, state$);
+            }
+          }
+        }
+
         const output$ = 'dependencies' in options
           ? epic(action$, state$, options.dependencies)
           : epic(action$, state$);


danielecr:
I can do it by:

```
const getJSON = (state_val, action_val, url) => ajax({url, headers: { "Authorization": "Bearer "+ state_val, "X-Someheader": action_val}, responseType: "json")

const epicMiddleware = createEpicMiddleware({
  dependencies: { getJSON }
});

```
and it is almost the same and as usable as well

jayphelps:
Mobile right now so can’t type full example but how I would do it is inject the dependency inside a custom root epic that then composes over the result of combineEpics (your old root epic). I’ll try and remember tomorrow to respond with example code but ping me if I forget and it’s not already clear. 🤙

jayphelps:
Also—could you provide a usage example where you’re providing the header and then using it with your proposed changes? That way I can I provide an equivalent.

eyal0803:
Here's an example of an epic that accesses the store, pulls the user token, and provides it to a function that then sends the AJAX request:
```js
const checkCodeEpic = (action$, state$) =>
  action$.ofType(AuthActionTypes.CHECK_CODE).pipe(
    switchMap(({ payload }) =>
      checkCode(payload, state$.value.auth.token).pipe(
        map(user => {
          if (user) {
            return AuthActions.checkCodeFulfilled(user)
          } else {
            return AuthActions.checkCodeRejected(true)
          }
        }),
        catchError(error => of(AuthActions.checkCodeRejected(error)))
      )
    )
  )
```
`checkCode` receives a user-provided `code` and a JWT which was stored at login time.

ksaldana1:
I don't know the details of your app, so I can only give some vague input. 

Why does your user access token need to be on your Redux state tree? I understand the idea of centralizing your "state", but it's not like you are ever rendering that access token in your React app. The only reason I can really think of is using it as a round-a-bout way to re-render when the access token changes.

In most apps I've worked in before, access token storage/generation/serialization is the responsibility of a higher-level "client". You should centralize this logic in a higher abstraction and inject that as a dependency. You may have methods on the client that allow you to interact with it (`setToken`) or whatever, but for the most part low-level token handling should be abstracted away.

jayphelps:
@eyal0803 Thanks of the code, though unfortunately I meant an example using the proposed changes here in the PR, using the injected `get` dependency that handles the auth for you inside an epic or two so that I can see what exactly you're looking for.

To be frank, it's unlikely we'll adopt something similar to this PR, but you can accomplish the same thing by composing inside a custom root epic:

```js
const rootEpic = (action$, state$, dependencies) => {
  const innerRootEpic = combineEpics(...epics);

  for (const dependency in dependencies) {
    if (dependencies.hasOwnProperty(dependency)) {
      const dependencyValue = dependencies[dependency];

      if (typeof dependencyValue === 'function') {
        dependencies[dependency] = () =>
          dependencyValue(action$, state$);
      }
    }
  }

  return innerRootEpic(action$, state$, dependencies);
};
```

FWIW I'm not sure this pattern of using dependencies for this sort of thing is ideal, but I'd need to see an actual usage example (or two) to provide alternatives.

jayphelps:
@eyal0803 Hey there! Did the pattern I mentioned about suit your needs?

jayphelps:
Closing, but lmk if you still need more help!

PR #604, opened 2019-01-29
Update typings to prevent TS error when using StateObservable (#1)

anirudhvarma12:
Update typings to prevent TS error when using StateObservable

Updates typings to prevent the following TS error when using an epic with StateObservable.

```
Argument of type 'Epic<SdAction, any, IAppState, any>' 
is not assignable to parameter of type 'Epic<Action, Action, void, any>'.
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -33,11 +33,13 @@ export declare class StateObservable<S> extends Observable<S> {
   value: S
 }
 
-export declare interface Epic<Input extends Action = any, Output extends Input = Input, State = any, Dependencies = any> {
+type State = any;
+
+export declare interface Epic<Input extends Action = any, Output extends Input = Input, State = State, Dependencies = any> {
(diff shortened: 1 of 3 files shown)


index.d.ts
```diff
@@ -33,21 +33,23 @@ export declare class StateObservable<S> extends Observable<S> {
   value: S
 }
 
-export declare interface Epic<Input extends Action = any, Output extends Input = Input, State = any, Dependencies = any> {
+type State = any;
```
evertbouw: This name is being shadowed now. Do we need to declare this separately? I think I'd prefer to use `any` directly.

anirudhvarma12: I thought it was better to be explicit that the parameter is for State, hence the type. I can revert this if required

jayphelps: If you're okay with it, I think we should use `any` explicitly rather than an alias of `type State = any`. Cool?

PR #610, opened 2019-02-12
[RFC] Refactor navigation example to use TestScheduler for epic tests

tanem:
I'm working on an app that uses `redux-observable`, and given everything I've read so far, it seems [the preferred method of testing epics is via the RxJS TestScheduler](https://redux-observable.js.org/docs/recipes/WritingTests.html).

It would be great if the examples contained in this repo demonstrated testing epics in this fashion. I can see this was attempted some time ago in #225, however that PR appears to have fizzled out due to difficulties testing `delay()`, and TestScheduler not being available back then.

This patch is intended as a conversation starter:

- Is the approach taken in [`adminAccess.test.js`](https://github.com/redux-observable/redux-observable/compare/master...tanem:refactor-navigation-example?expand=1#diff-d3be0e581f3bcf863f2673a98a1b5825) reasonable?
- If so, should we test the other example epics in this way? (I'd be keen to complete this work 🖐 ).

Note that the diff is quite large. Most of the noise is due the example app being created with create-react-app and updated to use newer library versions, similar to what was done in #578. I've tried to avoid going too far with this at this stage, since the focus of the PR is on the test approach, but it would make sense to do this properly if we decide to forge ahead with things I'm guessing.


examples/navigation/src/epics/adminAccess.test.js
```diff
@@ -0,0 +1,26 @@
+import { push } from 'react-router-redux'
+import { TestScheduler } from 'rxjs/testing'
+import { accessDenied, CHECKED_ADMIN_ACCESS } from '../actions'
+import adminAccessEpic from './adminAccess'
+
+describe('adminAccess epic', () => {
+  it('creates an accessDenied action first, then a route change action', done => {
+    const testScheduler = new TestScheduler((actual, expected) => {
+      expect(actual).toEqual(expected)
+      done()
```
jayphelps: Did you find his was necessary? It shouldn’t be.

jayphelps:
Yay! THANK YOU! 🙏

I’m still reviewing, will let you know.

examples/navigation/src/epics/adminAccess.test.js
jayphelps: That is, TestScheduler usage should run entirely synchronously since time is virtualized by it.

tanem: Oops, good spot, have removed it 👁

jayphelps:
Thank you again! <3

tanem:
Oh, thanks @jayphelps, didn't expect that 😅I'll get those other tests and library updates sorted in the next while too.

PR #625, opened 2019-03-21
fix(createEpicMiddleware): Don't share a scheduler queue with anyone else's RxJS code

### LINKED ISSUE #624: Epics miss action emitted after `epicMiddleware.run()` depending on environment
Depending on the environment when a store gets created and `epicMiddleware.run()` gets called, an action dispatched immediately after will get missed by the epics.  This seems to happen when `epicMiddleware.run()` occurs somewhere as a side effect of an observable using the queue scheduler (possibly others)

**If the current behavior is a bug, please provide the steps to reproduce and a minimal demo of the problem using JSBin, StackBlitz, or similar.**

https://stackblitz.com/edit/redux-observable-playground-emg4do?file=index.js

'TEST ACTION' will always get emitted to the epics regardless of where it's dispatched, as long as it's after `epicMiddleware.run()`. The programmer initializing the redux store and epic middleware should not have to be conscious of the code calling surrounding it.

Happens with latest version. Bug found in an electron app, reproduced in chromium (stackblitz)

jayphelps:
fixes #624

Since we use the same queueScheduler that other people's app code might also use, we would get different behavior if the rootEpic is passed to `epicMiddlweare.run(rootEpic)` inside an already running queueScheduler task. Before this PR we incorrectly relief on the fact that almost always this wouldn't happen, but it can. If it does, redux-observable won't subscribe to the rootEpic soon enough--it'll be deferred until the current queue task returns, potentially missing actions.

RxJS doesn't currently publicly export QueueScheduler constructor nor QueueAction, so we reach in using arguably private APIs.

@benlesh this is doing naughty things, reaching into (currently) private APIs. If you have a moment to give this a quick once over, lmk if you have alternative suggestions. If no time, no worries.

### DIFF
--- a/test/createEpicMiddleware-spec.js
+++ b/test/createEpicMiddleware-spec.js
@@ -5,7 +5,7 @@ import sinon from 'sinon';
 import { createStore, applyMiddleware } from 'redux';
 import { createEpicMiddleware, combineEpics, ActionsObservable, StateObservable, ofType } from '../';
 import { resetDeprecationsSeen } from '../lib/cjs/utils/console';
-import { of, empty, merge } from 'rxjs';
+import { of, empty, merge, queueScheduler } from 'rxjs';
 import { mapTo, filter, map, mergeMap, startWith, ignoreElements, distinctUntilChanged } from 'rxjs/operators';
 import { initAction } from './initAction';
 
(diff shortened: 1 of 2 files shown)


test/createEpicMiddleware-spec.js
```diff
@@ -414,4 +414,40 @@ describe('createEpicMiddleware', () => {
...
+
+      expect(store.getState()).to.deep.equal([
+        initAction,
+        { type: 'ACTION_1' },
+        { type: 'ACTION_2' },
+        { type: 'ACTION_3' },
+        { type: 'ACTION_4' }
+      ]);
+
+      done();
```
jayphelps: This isn't really "async" in the sense of needing this `done()` but I prefer to have it regardless in case of later refactors later that might cause the test to pass without actually testing any code.

jayphelps:
Landed in v1.1.0

PR #627, opened 2019-03-24
feat(createEpicMiddleware): Add createEpicMiddleware option gatekeep to swallow/emit actions only from epics

### LINKED ISSUE #626: Allow swallowing actions, emitting only epic-produced actions

Feature

`redux-observable` explicitly states that epic middleware is only ran _after_ the action has already passed through the proceeding middleware and state is changed.

There should be a mechanism (option) that allows actions to run through epics _before_ it reaches the reducers and subsequent middleware.

**Why**

Actions could come from untrusted sources (such as http or a socket) and have to be validated before they hit the resucers and change state. This change would allow the epics to validate (and swallow) actions depending on their validity.

halfmatthalfcat:
Fixes #626 

Adds the `createEpicMiddleware` option `gatekeep`, which forces actions through the current epic before hitting subsequent middleware and the reducers. Only actions produced by the epics will be passed through and processed accordingly. Any actions not produced by epics are effectively swallowed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ temp
 dist
 _book
 typings
+.idea
+yarn.lock
--- a/index.d.ts
+++ b/index.d.ts
@@ -43,6 +43,7 @@ export interface EpicMiddleware<T extends Action, O extends T = T, S = void, D =
 
 interface Options<D = any> {
   dependencies?: D;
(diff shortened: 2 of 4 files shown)


jayphelps:
https://github.com/redux-observable/redux-observable/issues/626#issuecomment-476381442

PR #628, opened 2019-04-04
Feature proposal: createDependencies option for createEpicMiddleware

MeLlamoPablo:

Hello!

When developing with `redux-observable`, we end up repeating the common pattern of passing `state$` to any dependency. For example, we have a `request` dependency which is a wrapper around `rxjs/ajax`. We pass `state$` to that function on every single call so that it can take the API token from the state and update the request headers accordingly.

My proposition is to introduce an option for `createEpicMiddleware` called `createDependencies` which must be a function that accepts the `state$` and returns the actual dependencies (the same dependencies that you'd pass to the `dependencies` option). This would avoid repeating code, as we wouldn't need manually to pass the `state$` to every single dependency.

If you like this change and will consider merging it, I suggest that we discuss whether or not `action$` should be passed too. I believe that it shouldn't, because I can't find any use cases for it, and might enable some anti-patterns. However, if there are legitimate use cases for it, implementing it in the future would break backward compatibility. So I think that this is an important discussion to have.

By the way, I regret that I committed these changes before realizing that #598 exists. I tried to search for this exact feature but couldn't find it. In [this comment](https://github.com/redux-observable/redux-observable/pull/598#issuecomment-453654980) you mention an alternative, but I think that the code example is hard to understand and honestly it feels a bit "hacky". I haven't tried it, but in TypeScript, it seems to me that we'd have to either use `any` or define some complex types.

If you need some examples of use cases, you can take a look at the commit f4dd2e042c1dcd579b738484d9f7d358bef782a5.

Thanks for your attention.

### DIFF
--- a/docs/api/createEpicMiddleware.md
+++ b/docs/api/createEpicMiddleware.md
@@ -16,7 +16,7 @@
 ### redux/configureStore.js
 
 ```js
-import { createStore, applyMiddleware, compose } from 'redux';
+import { createStore, applyMiddleware } from 'redux';
 import { createEpicMiddleware } from 'redux-observable';
 import { rootEpic, rootReducer } from './modules/root';
 
--- a/docs/api/createEpicMiddleware.md
+++ b/docs/api/createEpicMiddleware.md
@@ -5,7 +5,8 @@
 #### Arguments
 
 1. *`[options: Object]`*: The optional configuration. Options:
-    * *`dependencies`*: If given, it will be injected as the 3rd argument to all epics.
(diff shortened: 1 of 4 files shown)


jayphelps:
Hey there! I've been on sabbatical (still am) so I haven't done much open source lately but wanted to at the very least acknowledge you!

Thanks so much for putting the time into this. Indeed I think the idea is similar to #598. I'd love to hear from more folks in the community. I think it's a bit too much opinionated of a pattern for us to encourage, but I'm open to discussions.

evertbouw:

PR #634, opened 2019-04-27
Improve readability of ofType operator

mikeheddes:

### DIFF
--- a/src/operators.js
+++ b/src/operators.js
@@ -1,21 +1,9 @@
 import { filter } from 'rxjs/operators';
 
-const keyHasType = (type, key) => {
+const matchesType = type => key => {
   return type === key || typeof key === 'function' && type === key.toString();
 };
 
-export const ofType = (...keys) => (source) => source.pipe(
-  filter(({ type }) => {
-    const len = keys.length;
-    if (len === 1) {
-      return keyHasType(type, keys[0]);
-    } else {
-      for (let i = 0; i < len; i++) {
-        if (keyHasType(type, keys[i])) {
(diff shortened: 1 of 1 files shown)


src/operators.js
```diff
@@ -1,21 +1,9 @@
...
-      for (let i = 0; i < len; i++) {
-        if (keyHasType(type, keys[i])) {
-          return true;
-        }
-      }
-    }
-    return false;
-  })
+export const ofType = (...keys) => source => source.pipe(
+  filter(({ type }) => keys.some(matchesType(type)))
```
mikeheddes: It could also be `keys.includes` but according to [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) that is not supported on all browsers. I'm not sure if IE must be supported.  Another option could be to use `keys.indexOf(...) !== -1` but that doesn't read like English where as `keys.some(matchesType(type)` does (sort of).

evertbouw:
The same change was proposed here https://github.com/redux-observable/redux-observable/pull/361 and the conclusion was that performance is more important than readability. It seems that v8 has optimised array.some a lot since then, performance looks on par with for loops.

@rgbkrk @jayphelps is it time for this now?

mikeheddes:
I should have made a comment on the already existing PR but I didn't went through the open PR's. Sorry for that.

So I think this PR can be closed and if needed continue the conversation in #361?

jayphelps:
Thanks @mikeheddes! <3

PR #663, opened 2019-09-17
fix(build): Don't mix ES Module and CommonJS syntax

pat-son:

It is bad form to mix ES Module `import`s and CommonJS `require`s. It also causes problems when using Rollup, because it skips CommonJS module resolution when it encounters ES Module syntax. [See here](https://github.com/rollup/rollup/issues/1058#issuecomment-254187433).


jayphelps:
Awesome! Thank you! 🎉 

For context, it was done this way during a time when ESM tree shaking was not yet as widely used and we wanted to exclude it from production builds. I think today tooling is in a good enough shape that if you care about bundle size enough to want the util helper excluded, using ESM tree shaking is the answer.

jayphelps:
Released as 1.2.0

PR #672, opened 2019-10-11
Convert project to Typescript and add es2015 build target

### LINKED ISSUE #644: Store creation broken with Stencil One
I'm reporting a bug after waiting answers for this stackoverflow [post](https://stackoverflow.com/questions/56350558/cant-use-redux-observable-with-stencilone-class-constructor-observable-cann).

Any project with stencil@1.0.0-beta.x and redux-observable should fail at runtime when creating store with this error :
```
TypeError: Class constructor Observable cannot be invoked without 'new'
    at new ActionsObservable (chunk-f18fe9c6.js:3744)
    at epicMiddleware (chunk-f18fe9c6.js:3830)
    at chunk-f18fe9c6.js:4579
    at Array.map (<anonymous>)
    at chunk-f18fe9c6.js:4578
    at createStore (chunk-f18fe9c6.js:4043)
    at chunk-f18fe9c6.js:4716
```

I expected redux-observable to work with Stencil One.

Npm packages :

- @Stencil/core : 1.0.0-beta.7
- @Stencil/redux : 0.1.1
- redux : 4.0.1
- redux-actions : 2.6.5
- redux-observable : 1.1.0

Node version 12.1.0
Npm version 6.9.0
Browser : Chrome Version 70.0.3538.77
OS : Ubuntu 18.04.2

Please look at the stackoverflow post for more informations.

csvn:

This PR tries to fix #644, which happens when native classes are used via `RxJS` from the `es2015 package.json` property. This library then crashes due to the transformed classes are trying to call class constructors instead of using `new`. `RxJS` has `main`, `module` and `es2015` fields in their `package.json`. This PR adds `es2015` to this library as well, so native classes are used at the same time if the same `package.json` fields are used.

### Updated implementation notes

- All source/test files have been migrated to Typescript
- Typescript version has been updated to `^3.6.4` from `^2.1.5`
- Uses `strict` and `importHelpers` in `tsconfig.json`
- Linting is still done via `eslint`, but `@typescript-eslint` is used to add Typescript support as well as a few recommended Typescript rules
- Files are transpiled via Typescript instead of babel/gulp
- Bundles are still built with Webpack, but now uses `ts-loader` instead of `babel-loader`

----------------------------------------------

### Original implementation notes (babel)

Babel is not well suited to transform to a specific language level, so transforming to ~es2015 is kind of hacky in this implementation. The best solution would be to instead compile this library as `RxJS` does; via Typescript. I did not want to do that in this PR without checking if there was any interest in this. There are several benefits to using Typescript instead of babel:

* `RxJS` and `redux-observable` are used together, using similar tooling might be beneficial
* It would be easier to generate `es2015` the same way as `RxJS`
* Less extraneous runtime helpers (e.g. `_createClass`, `_inherits`, `_classCallCheck`) that are repeated throughout the compiled `redux-observable` code (`RxJS` uses `tslib`, so all helper functions are imported)
* If files are changed to Typescript, `index.d.ts` would no longer need to be maintained

The `lib/es2015` folder is ~4KB smaller (11KB -> 7KB) on my machine, and much of this can be attributed to most runtime helpers no longer being necessary. If this project was compiled with TS and `importHelpers`, the `cjs` and `esm` builds would get smaller too.

If converting to `.ts`, or just compiling `.js` with Typescript would be agreeable, let me know and I'll update the PR.

> Updated 1: New implementation in Typescript instead of babel


jayphelps:
This is AWESOME!! and I'm definitely down for porting to TypeScript--it's long overdue tbh Did you volunteer? If you don't have cycles, I should be able to do it soon. If it's easier for a first pass, feel free to skip strict options like no implicit any, non-nullables, etc and I can come in and fix it if there are cases where we're doing things that might require heavy refactoring to make type safe/not awkward.

Thank you!! <3

csvn:
@jayphelps We use Typescript with all our projects, so I wouldn't mind taking a stab at this!

configs/tsconfig.defaults.json
```diff
@@ -0,0 +1,17 @@
+{
+  "compileOnSave": false,
+  "compilerOptions": {
+    "target": "es5",
+    "module": "es2015",
+    "moduleResolution": "node",
+    "outDir": "dist",
+    "importHelpers": true,
```
csvn: I have not added `tslib` as a peerDependency, though `rxjs` has it as a dependency and since `rxjs` is a peerDependency it should still work. Though, it might still be good to be explicit and add it?

csvn:
There was a CI error due to a some changes made for `@typescript-eslint` at version `2.4.0`. It should be fixed in next version (https://github.com/typescript-eslint/typescript-eslint/pull/1088). The `PARSER_NO_WATCH=true` environment variable should be able to be removed then.

test/typings.ts
```diff
@@ -1,5 +1,5 @@
 import { expect } from 'chai';
-import { createStore, applyMiddleware, MiddlewareAPI, Action } from 'redux';
+import { createStore, applyMiddleware, Action } from 'redux';
 import { Observable, asapScheduler, of } from 'rxjs';
```
jayphelps: Do we still need this test/typings.ts file? I don't think we would, since the real tests are now in TypeScript.

src/StateObservable.ts
```diff
@@ -1,7 +1,13 @@
-import { Observable, Subject } from 'rxjs';
+import { Observable, Subject, Subscription } from 'rxjs';
 
-export class StateObservable extends Observable {
-  constructor(stateSubject, initialState) {
+export class StateObservable<S> extends Observable<S> {
+  value: S;
+  // TODO: This property is never accessed, should it be removed?
+  // @ts-ignore
+  private __subscription: Subscription;
```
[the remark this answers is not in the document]
jayphelps: It can be removed

src/combineEpics.ts
```diff
@@ -1,10 +1,17 @@
...
 /**
   Merges all epics into a single one.
  */
-export const combineEpics = (...epics) => {
-  const merger = (...args) => merge(
+export function combineEpics<T extends Action, O extends T = T, S = void, D = any>(...epics: Epic<T, O, S, D>[]): Epic<T, O, S, D>;
+// TODO: Maybe we should make sure `E` is callable? (`E extends () => ...`)
+export function combineEpics<E>(...epics: E[]): E;
+// TODO: This typing seems brittle, is `any` really needed?
+export function combineEpics(...epics: any[]): any;
```
jayphelps: Not sure. We could remove it, and see if anyone complains.

src/combineEpics.ts
```diff
@@ -1,10 +1,17 @@
...
+import { Epic } from './epic';
 
 /**
   Merges all epics into a single one.
  */
-export const combineEpics = (...epics) => {
-  const merger = (...args) => merge(
+export function combineEpics<T extends Action, O extends T = T, S = void, D = any>(...epics: Epic<T, O, S, D>[]): Epic<T, O, S, D>;
+// TODO: Maybe we should make sure `E` is callable? (`E extends () => ...`)
+export function combineEpics<E>(...epics: E[]): E;
```
jayphelps: That would be an improvement, but feel free to punt on that for this PR if you'd like. Either way.

src/createEpicMiddleware.ts
```diff
@@ -1,44 +1,55 @@
...
       warn('this middleware is already associated with a store. createEpicMiddleware should be called for every store.\n\nLearn more: https://goo.gl/2GQ7Da');
     }
     store = _store;
-    const actionSubject$ = new Subject().pipe(
+    // `.pipe()` transforms typing from `Subject<T>` to `Observable<T>`
+    const actionSubject$ = new Subject<T>().pipe(
       observeOn(uniqueQueueScheduler)
-    );
-    const stateSubject$ = new Subject().pipe(
+    ) as any as Subject<T>;
```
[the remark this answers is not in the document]
jayphelps: We could refactor out a separate "actionSubjectOutput$` or something, but this is good enough for now.

src/createEpicMiddleware.ts
```diff
@@ -1,44 +1,55 @@
...
     const action$ = new ActionsObservable(actionSubject$);
     const state$ = new StateObservable(stateSubject$, store.getState());
 
     const result$ = epic$.pipe(
       map(epic => {
         const output$ = 'dependencies' in options
-          ? epic(action$, state$, options.dependencies)
-          : epic(action$, state$);
+          ? epic(action$, state$, options.dependencies!)
+          : (epic as any)(action$, state$);
```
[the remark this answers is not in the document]
jayphelps: That's a good point. I'm not sure what the ideal is here, because while it _could_ be optional, if someone is using it, forcing them to do undefined checks every time is a real pain. Maybe keep the types as-is and change our code here to always pass dependencies, even if it's undefined. The function arity will change, but that's ok because I'm gonna release TS stuff as a major version bump anyway, so breaking changes are possible.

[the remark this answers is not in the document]
jayphelps: What do you think though? Disagree? Alternatives?

src/operators.ts
```diff
@@ -1,11 +1,14 @@
+import { Action } from 'redux';
+import { Observable } from 'rxjs';
 import { filter } from 'rxjs/operators';
 
-const keyHasType = (type, key) => {
+const keyHasType = (type: unknown, key: unknown) => {
```
jayphelps: Nice 👍

src/createEpicMiddleware.ts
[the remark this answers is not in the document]
csvn: I think I'll leave this for now. It's not pretty, but it works

src/StateObservable.ts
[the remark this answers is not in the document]
csvn: Removed.

src/createEpicMiddleware.ts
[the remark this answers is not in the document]
csvn: I had to leave a non-null assertion, since `D | undefined` is not assignable to `D`. There are a couple of places the generics makes the code a little weird, but this can be improved over time.

```ts
const output$ = epic(action$, state$, options.dependencies!);
```

src/combineEpics.ts
csvn: Since you mentioned this will make semver major, I think it's a good idea to add a minimal constraint here: `<E extends () => Observable<any>>`

jayphelps: Sounds good. I'm not entirely sure why this overload is needed. It might be, or it might only exist as a remnant over the years. It would seem that the first signature, with Epic<>, would work in any valid case wouldn't it? The only argument I can see is that the second overload is far less strict, so it would accept functions that aren't yet typed 100% as epics, which I'm not sure is a good thing. If someone wants that, they can always cast to any. Thoughts?

jayphelps:
The failing test `const rootEpic = combineEpics(() => 'anonymous', epic2);` is invalid, the epic shouldn't return a string. (this code was already there, not something you added)

csvn:
Yeah, forgot to run tests before I made the commit. I'll update the PR shortly.

src/combineEpics.ts
csvn: I checked the blame for the line. It was added [in this PR](https://github.com/redux-observable/redux-observable/commit/88c0bf9e9bd44feeed1af18ec7e5639890400e72) that mentions dependency injection, but this seems to be before the `Dependencies` generic was added to the `Epic` interface. Could be this is not necessary anymore due to being able to use:
```ts
createEpicMiddleware({ dependencies: myDeps });
```

I'll remove this overload too. I can always revert otherwise.

csvn: I removed all overloads for `combineEpics` in last commit.

csvn:
Again I forgot to fix run tests before comitting. I'll try to fix it tonight. 😅

configs/tsconfig.defaults.json
csvn: I added `tslib` as a peerDependency. I used the same version as RxJS for now.

csvn:
Tests are passing again, and I think I've addressed your comments. I decided not to change all the tests that breaks the typing. Some are intentionally verifying that error is thrown if epic does not return a stream. Some tests are just simpler of they do not send exact correct types to an epic. I can of course adjust this if you think that's appropriate.

jayphelps:
All LGTM. If anyone has objections or changes, lmk, but I'm gonna merge for now.

Won't be releasing this immediately, as I'd like to do this under 2.0.0 major version bump and there are some other things that are almost ready for that.

If anyone is interested in alpha testing a build of this, I can publish as a special tag to npm, just lmk. Otherwise I'll chime in when I've done that anyway, probably in a couple days.

jayphelps:
Thanks again @csvn 🎉 !

csvn:
Glad I could help @jayphelps! 😀

csvn:
Hey @jayphelps, sorry to bother you. I'd be interested to try out an alpha of the 2.0.0 version if possible, unless the latest release is close. No pressure though 🙂

jayphelps:
@csvn appreciate the apology, but it's perfectly all right! Indeed wanted you to let me know if/when a pre-release would be helpful. I'll try and get one in the next day or two and will update here.

jayphelps:
I've cut a pre-release `2.0.0-alpha.0` or the `next` tag (which will be updated to whatever the latest pre-release is.

csvn:
Cool, thanks!

PR #678, opened 2019-10-19
chore(build): add package-lock.json,  add tslib to devDeps

jayphelps:
* ~~remove extraneous non-null assertion (TypeScript complained about this)~~ old code was correct
* add package-lock.json
* add tslib to devDeps (build failed because peerDeps aren't installed by default)

Cc/ @csvn

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,0 @@
-package-lock=false
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,14293 @@
+{
+  "name": "redux-observable",
+  "version": "1.2.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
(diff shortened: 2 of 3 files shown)


jayphelps:
wtf local typescript says it's not needed, Travis's TS says it is. 😕 

Investigating...

jayphelps:
Fixed. Not sure what happened, but I assume my local install resolved TS to an older version that wasn't as strict as newer ones, because we didn't have a lock file. The non-null assertion seems correct, and TS now is happy with it, so leaving it.

csvn:
Ahh, right, forgot about adding `tslib` to `devDependencies` too 😅

I noticed the missing lockfile, but I didnt know if that was intentional or not. If VSCode is used, local TS error may also differ if built-in or local version of Typescript is used.

Issue #622, opened 2019-03-01
Improve `ofType` type declaration for TS >3.3


tjfryan:
**Do you want to request a *feature* or report a *bug*?**  

I guess this would qualify as a feature request.

**What is the current behavior?**

When using redux-observable in TypeScript, providing a key or keys to `ActionsObservable.prototype.ofType` doesn't refine the type of the actions emitted and often requires a redundant `action.type` check in order to maintain type correctness.

**What is the expected behavior?**

[...]

PR #681, opened 2019-10-19
Type inference for ofType, removal of ActionsObservable in favor of just Observable

jayphelps:
Heavily inspired by examples that @thorn0 @tjfryan did in #622

***

For v2.0.0 I'd like to finally have ofType infer the TypeScript types automatically, so that you don't need to pass in the generic param or define your filter action(s) as the argument type of the next operator. This requires some voodoo and a new-ish version of TypeScript, but I'm not yet sure which minimum version. I'll have to try them until it breaks, before merging this, then I can document it in the changelog/docs.

At the same time, I want to get away from ActionsObservable because in the pipeable operator world, there's no need for it and it just adds friction/learning.

The primary reason StateObservable is not BehaviorSubject is because we want to expose `state$.value` but _we do not want to expose `state$.next(something)` because that would be a huge footgun that wouldn't do what you might think--no reason to do it inside your epics.

***

Have objections or other thoughts? Please lmk! We'll likely do the normal pre-alpha, alpha, beta, rc, etc cycle for 2.0, but depending on how smooth it goes, how many changes there are (don't expect _too many_) we might skip a step or two of those.


kwonoj:
🚢

evertbouw:
I don't think it's necessary to deprecate ActionObservable before removing it. Changing the ofType method to an imported function is not a huge change and probably the only thing users will have to do, and they may have already done it.

src/createEpicMiddleware.ts
```diff
@@ -42,7 +41,7 @@ export function createEpicMiddleware<T extends Action, O extends T = T, S = void
     const stateSubject$ = new Subject<S>().pipe(
       observeOn(uniqueQueueScheduler)
     ) as any as Subject<S>;
```
csvn: Nit: ` as any as Subject<S>` can be removed, now that a regular `Observable` is used instead of `ActionsObservable`.

I love this PR, as I think it simplifies both usage and the code quite a bit. Sorry if I just jumped in with this review.

jayphelps: Wonderful! Thank you for pointing this out.

jayphelps: Turns out it couldn't just be removed as-is because while observeOn() uses RxJS's internal lift() function so it does return a Subject, the type definition does not reflect that with pipe() so from the type system point of view it's just an Observable and doesn't have .next() nor .asObservable(). No biggie as I was able to refactor the code in another way to avoid the casts.

jayphelps:
Thanks for all the reviews! I've cut a pre-release `2.0.0-alpha.0` or the `next` tag (which will be updated to whatever the latest pre-release is.

csvn:
@jayphelps I noticed the release commit 9a85f64a7768e784ec0da7c5a98ec1b0c5c603b4, but no publish seems to have been done to NPM (https://www.npmjs.com/package/redux-observable?activeTab=versions). Maybe there's some issue with the publishing pipeline. Should I create an issue for this?

jayphelps:
Whoa. I know 100% for sure the 2.0 alpha was released and tagged as next. It’s been installed and used in several apps that I’m aware of. Not sure what happened. I’ll look into it ASAP likely this weekend. Thanks for the heads up! Feel free to make a ticket if you’d like to subscribe to updates

jayphelps:
Proof I’m not crazy.
A370ED88-F3F5-4D76-9D86-142AB2CF758E

evertbouw:
`npm info redux-observable` shows
```
dist-tags:
latest: 1.2.0        next: 2.0.0-alpha.0  
```

jayphelps:
Phew! Thanks for checking that @evertbouw it must just be a UI bug or DB issue. I’m not able to try to install it right now, but anyone should let us know if you can’t install.

csvn:
That's weird... I didn't run `npm info redux-observable`, I just checked the NPM page, but it's still not showing the new alpha version 🤔

image

Thanks a lot, sorry for the noise!

israelidanny:
I'm trying to play around with this but can't seem to get the type inference to work. Where do I supply my actions union type for the inference to work?

jayphelps:
@israelidanny Unfortunately this didn't actually make ofType infer the types correctly. I swear it worked when I PRd it, but apparently not. 😢 I haven't had time to prioritize revisiting this as I'm not working on any redux-observable codebases at the moment. If someone can get it to work, I'd super appreciate any suggestions or even a PR if they're feeling ambitious. Otherwise I'll for sure _eventually_ get to it, but can't say when.

kwonoj:
Coincidentally I just tried to update my code to use this and hitting same thing, thought I was doing something wrong. 😭

csvn:
@jayphelps @kwonoj @israelidanny We are able to use type inference in our code for `ofType`, but it requires some thought. What I think is the issue above is that Typescript is widening the types. For example, `'LOGIN_REQUEST'` becomes `string` instead of the literal type `'LOGIN_REQUEST'`.

The below is a working example of inference for `ofType` by using `as const` to avoid Typescript widening the type from the specific `enum` value to `ActionType`. The example works the same if replacing the `enum` with hard-coded string values.

```ts
import { Epic, ofType } from 'redux-observable';
import { tap, ignoreElements } from 'rxjs/operators';


enum ActionType {
  Update = 'COUNT_UPDATE',
  Reset = 'COUNT_RESET'
}

type State = any;
type Action = ReturnType<typeof setCount> | ReturnType<typeof resetCount>;

const setCount = (count: number) => ({
  type: ActionType.Update as const, // 1
  payload: count
});
const resetCount = () => ({
  type: ActionType.Reset as const // 1
});

export const testEpic: Epic<Action, Action, State, undefined> = action$ => action$.pipe(
  ofType(ActionType.Update as const), // 1
  tap(({ payload }) => console.log(payload + 1)),
  ignoreElements()
);
```

It might be possible to improve the typing so you don't need `ofType('<TYPE>' as const`)`, but I have no idea where to start. I find some of the more intricate matters with generics and Typescript a bit confusing 😄

danmcdaid:
@csvn Thanks for the above comment.  Was scratching my head with this.  Since your comment, do you have any further understanding of why the enum type does not get narrowed? 🤔

PR #720, opened 2020-03-20
docs(troubleshooting): fix how Arrow Functions works

piecioshka:

### DIFF
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -66,19 +66,20 @@ If you are organizing your epics into a class. (E.g. in order to benefit from An
 
 ```typescript
 class TooFancy {
-  constructor(private somethingInjected:SomethingInjected)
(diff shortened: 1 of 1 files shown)


docs/Troubleshooting.md
```diff
@@ -66,19 +66,20 @@ If you are organizing your epics into a class. (E.g. in order to benefit from An
...
 }
 ```
+
 follow the docs and:
 
 ```typescript
 class TooFancy {
-  constructor(private somethingInjected:SomethingInjected)
-  checkAutoLogin =  (action$: Observable<IPayloadAction>) => {
-    console.log(this); // YOu can access somethingInjected
```
jayphelps: Hmm I believe there is some confusion. I believe that section is saying you NEED to use arrow functions so that you close over the `this` of the class instance. The example is showing the before/after. It probably can be reworded to make that more clear though. Lmk if I'm misunderstanding.

PR #732, opened 2020-07-11
Fixing the Cancellation docs.

jun-sheaf:
The previously written example is inaccurate. For example, `incrementCounter()` could emit two <del>cancellation</del> actions, but placing `take(1)` after `map(() => incrementCounter())` would only allow one through.

Also fixed some grammatical mistakes.

### DIFF
--- a/docs/recipes/Cancellation.md
+++ b/docs/recipes/Cancellation.md
@@ -1,6 +1,6 @@
 # Cancellation
 
-Cancelling some async side effects is a common requirement of Epics. While there are several ways of doing this depending on your requirements, the most common way is to have your application dispatch a cancellation action and listen for it inside your Epic.
+Canceling some async side effects is a common requirement of Epics. While there are several ways of doing this, depending on your requirements, the most common way is to have your application dispatch a cancellation action and listen for it inside your Epic.
 
 This can be done with the [`takeUntil()`](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-takeUntil) RxJS operator:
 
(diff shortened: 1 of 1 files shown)


jayphelps:
While I do think putting the take(1) immediately after the ofType() probably makes it more clear, it actually makes no practical difference in this example. It would not allow two actions to emit. Can you clarify?

jun-sheaf:
I retracted the word `cancellation` from my comment since that was actually incorrect. But even with the retraction, since we are using `map`, necessarily there can only be one action emitted, so it is still not totally accurate.

In any case, if consider replace `map(() => incrementCounter())` with `mergeMap(() => of(incrementCounter(),  incrementCounter()))`. Then two increment actions should be emitted, but with the previous location of `take(1)`, this is not the case.

docs/recipes/Cancellation.md
```diff
@@ -1,6 +1,6 @@
 # Cancellation
 
-Cancelling some async side effects is a common requirement of Epics. While there are several ways of doing this depending on your requirements, the most common way is to have your application dispatch a cancellation action and listen for it inside your Epic.
+Canceling some async side effects is a common requirement of Epics. While there are several ways of doing this, depending on your requirements, the most common way is to have your application dispatch a cancellation action and listen for it inside your Epic.
```
jayphelps, line 3: If we change cancelling to canceling we need to do the spelling style change everywhere. There is at least one missed (I haven’t checked for others) the heading “Cancellation”

jayphelps: Ah the file itself too, maybe other places. Otherwise feel free to undo these changes as another option.

jayphelps: Also remember to change any links too. It’s unfortunate that changing the file will cause 404s for SEO. Wonder what could be done easily.

jun-sheaf: I'll revert those changes.

jun-sheaf: Actually, I'll take a better look at the docs and see if I can improve them. There seems to be a load of old Rxjs references and some outdated answers.

PR #733, opened 2020-07-12
Better Troubleshooting and Cancellation Docs

jun-sheaf:
Changes:

* All the links have been updated to the latest `rxjs` docs.
* Headings have been reformatted to be more standard.
* Fixed some grammar.
* Particular to `Troubleshooting.md`
   * Workarounds have been added for the `ofType` operator issue.
* Particular to `Cancellation.md`
   * Examples have been improved.
   * Remarks have been improved.


docs/Troubleshooting.md
```diff
@@ -25,32 +24,31 @@ import 'rxjs';
...
 ```js
 import 'rxjs/add/operator/switchMap';
 action$.ofType(...).switchMap(...);
 
 // OR
 
-import { switchMap } from 'rxjs/operator/switchMap';
-action$.ofType(...)::switchMap(...);
+import { switchMap } from 'rxjs/operators';
+action$.ofType(...).switchMap(...);
```
jayphelps, line 40: This was actually not a typo. It was just an example using the [bind operator](https://github.com/tc39/proposal-bind-operator) proposal. But that proposal is unlikely to ever make it, so removing that example is probably a good idea to remove it.

```suggestion
import { switchMap } from 'rxjs/operators';
import { ofType } from 'redux-observable';
// If you are using a compiler with pipeline operator support (non-standard)
action$ |> ofType(...) |> switchMap(...);
```

jayphelps: Alternatively we could remove this suggestion altogether, since the pipeline operator is also non-standard still. Though there is more hope it might one day land.

docs/recipes/Cancellation.md
```diff
@@ -2,59 +2,57 @@
...
 
-This can be done with the [`takeUntil()`](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-takeUntil) RxJS operator:
-
+This can be done with the [`takeUntil`](https://rxjs-dev.firebaseapp.com/api/operators/takeUntil) RxJS operator:
 ```js
-import { ajax } from 'rxjs/ajax';
-
 const fetchUserEpic = action$ => action$.pipe(
-  ofType(FETCH_USER),
+ ofType(FETCH_USER),
```
jayphelps, line 8: Looks like accidentally removed a single space before this.
```suggestion
  ofType(FETCH_USER),
```

docs/recipes/Cancellation.md
```diff
@@ -2,59 +2,57 @@
...
   mergeMap(action => ajax.getJSON(`/api/users/${action.payload}`).pipe(
-    map(response => fetchUserFulfilled(response)),
-    takeUntil(action$.pipe(
-      ofType(FETCH_USER_CANCELLED)
-    ))
-  ))
+   takeUntil(action$.pipe(ofType(FETCH_USER_CANCELLED))),
+   map(response => fetchUserFulfilled(response))
```
jayphelps, line 11: Two space indention please.
```suggestion
    takeUntil(action$.pipe(ofType(FETCH_USER_CANCELLED))),
    map(response => fetchUserFulfilled(response))
```

docs/Troubleshooting.md
```diff
@@ -168,6 +170,23 @@ const epic = (action$: ActionsObservable<Actions>) =>
...
+
+import { OperatorFunction } from "rxjs";
+import { Action } from "redux";
+
+declare module "redux-observable" {
+ function ofType<T extends Action<unknown>, K extends string>(
+  ...key: K[]
+ ): OperatorFunction<T, Extract<T, { type: K }>>;
+}
+```
```
jayphelps, line 188: Oh it looks like perhaps your editor must have had 1 space indention. Can you go through and make sure they're all two spaces?

jun-sheaf:
Thanks for the review!

So I have made some amendments, but before I push them, there needs to be some discussion about the `Troubleshooting` guide. In my opinion, aside from migration and the typing issues, why should we include guides that are fundamentally from `RxJS`?

PR #734, opened 2020-07-23
Change state type param default to any

jun-sheaf:
The following changes were made:
* The type names have been changed to be semantic.
* `State` defaults to `any`.
* The deleterious method of grabbing `RxJS`'s `queueScheduler` constructor and action has been removed and replaced with a legitimate imports from `RxJS` internals.

### DIFF
--- a/src/createEpicMiddleware.ts
+++ b/src/createEpicMiddleware.ts
@@ -1,9 +1,9 @@
-import { Action as ReduxAction, Dispatch, Middleware, MiddlewareAPI } from 'redux';
-import { from, Subject, queueScheduler } from 'rxjs';
-import { map, mergeMap, observeOn, subscribeOn } from 'rxjs/operators';
-import { Epic } from './epic';
-import { StateObservable } from './StateObservable';
-import { warn } from './utils/console';
+import { Action as ReduxAction, Dispatch, Middleware, MiddlewareAPI } from "redux";
+import { from, Subject, queueScheduler } from "rxjs";
+import { map, mergeMap, observeOn, subscribeOn } from "rxjs/operators";
+import { Epic } from "./epic";
+import { StateObservable } from "./StateObservable";
+import { warn } from "./utils/console";
 
 interface Options<Dependencies = any> {
   dependencies?: Dependencies;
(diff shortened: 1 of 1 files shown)


src/createEpicMiddleware.ts
```diff
@@ -1,57 +1,52 @@
...
-  D = any
->(options: Options<D> = {}): EpicMiddleware<T, O, S, D> {
-  // This isn't great. RxJS doesn't publicly export the constructor for
-  // QueueScheduler nor QueueAction, so we reach in. We need to do this because
-  // we don't want our internal queuing mechanism to be on the same queue as any
-  // other RxJS code outside of redux-observable internals.
-  const QueueScheduler: any = queueScheduler.constructor;
-  const uniqueQueueScheduler: typeof queueScheduler = new QueueScheduler(
-    (queueScheduler as any).SchedulerAction
-  );
```
jayphelps, line 34: We didn’t use an internal import because those imports are not guaranteed to be stable between minor/patch releases because they are internal. Whereas, while dirty, the `queueScheduler.constructor` technique should always work as long as queueScheduler is exported, which if changed wouldn’t until a major version release.

jayphelps: The real solution is to have rxjs export the constructor publicly, but we just haven't got around to it haha

jun-sheaf: Ah, okay then. I'll revert this and perhaps reference this PR in case of future PRs.

jun-sheaf:
I'll remove the style changes. Those were a result of my editor; it wasn't intentional.

Also I've submitted the todos previously. I removed them because of the internal comment. Perhaps we should remove the "first timer" comment in the Issues template.

jayphelps:
> Also I've submitted the todos previously. I removed them because of the internal comment. Perhaps we should remove the "first timer" comment in the Issues template.

Ah I see what happened. It says you were a first time contributor since the previous PR you submitted wasn't merged.
image

Btw heads up that I changed the PR title because "See Description" would make it difficult to know what this PR is for later or if people are searching for it. Feel free to change it again if you have an alternative descriptive title!

jun-sheaf:
@jayphelps This request is ready to be merged 👍

jun-sheaf:
@jayphelps Could we get this approved?

src/createEpicMiddleware.ts
```diff
@@ -90,7 +87,7 @@ export function createEpicMiddleware<
         // updated before epics receive the action
         const result = next(action);
 
-        // It's important to update the state$ before we emit
+        // It"s important to update the state$ before we emit
```
jayphelps, line 90: Whoops

jayphelps:
@jun-sheaf would you mind removing the code style changes? e.g. remove the single-to-double quotes change, newlines, etc.

PR #742, opened 2021-03-16
fix(package.json): avoid tslib version issues

### LINKED ISSUE #740: tslib should be listed in dependencies instead of peerDependencies
As tslib is listed as peer dependencies,  it will be ensured that only one instance will be installed and shared by redux-observable and user-project, redux-observable will get an unexpected version of tslib at runtime, and may hit any breaking change of tslib.

csvn:
This PR fixes issues with NPM peer-dependency installation when the host project uses `tslib` >2.0.0.

image

Closes #740

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
   "peerDependencies": {
     "redux": ">=4 <5",
     "rxjs": ">=6.0.0-beta.0 <7",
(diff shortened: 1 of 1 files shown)


csvn:
No worries 😊

PR #743, opened 2021-04-08
docs: remove broken jsbin embeds

jayphelps:
These have been broken due to that I no longer pay for a jsbin pro account.


jayphelps:
btw publishing the docs is currently broken as-is due to some bug in one of the dependencies with the newer versions of Node. This solves it until I or someone else has time to move away from gitbook (or find an alternative workaround) https://flaviocopes.com/cb-apply-not-a-function/

PR #746, opened 2021-04-19
Setup GitHub actions

MaximeBernard:
I figured out it would be a good idea to use Github Actions:
- better integration with PR & Github UI overall
- one place to manage everything

We needed to fix that Travis CI anyway

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,7 @@
-website
-temp
 dist
+docs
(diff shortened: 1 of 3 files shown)


evertbouw (approved):

jayphelps (approved):
Seems legit. Thank you much for doing this!

PR #747, opened 2021-05-01
feat(rxjs7): add rxjs 7 compatibility

### LINKED ISSUE #735: Rxjs 7 compatibility
When bumping the dependency of RxJS to v6 I made the mistake of trying to predict the future and didn't allow v7. Should we make a 1.x release that does allow it, or make this part of 2.0?

eltiodelsofa:
BREAKING CHANGE: Uses typescript 4.2.2 and tslib 2.1.0 (same libs as rxjs 7). @typescript-eslint/eslint-plugin and @typescript-eslint/parser has been updated to 4.22.0 to run linter properly with new ts version, new troublesome linter rules added.

Closes #735

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -257,6 +257,11 @@
     "@typescript-eslint/unbound-method": "off",
     "@typescript-eslint/ban-ts-ignore": "off",
     "@typescript-eslint/no-empty-function": "off",
+    "@typescript-eslint/no-unsafe-assignment": "off",
+    "@typescript-eslint/no-unsafe-call": "off",
+    "@typescript-eslint/no-unsafe-member-access": "off",
+    "@typescript-eslint/no-unsafe-return": "off",
+    "@typescript-eslint/restrict-plus-operands": "off",
     // disallow unused variables/imports
     "@typescript-eslint/no-unused-vars": [2, {
       "vars": "all",
--- a/package.json
+++ b/package.json
@@ -66,16 +66,16 @@
   "homepage": "https://github.com/redux-observable/redux-observable#README.md",
   "peerDependencies": {
     "redux": ">=4 <5",
-    "rxjs": ">=6.0.0-beta.0 <7",
-    "tslib": "^1.9.0"
+    "rxjs": "^7.0.0",
+    "tslib": "~2.1.0"
   },
   "devDependencies": {
     "@types/chai": "^3.5.2",
     "@types/mocha": "^2.2.48",
     "@types/node": "^12.7.12",
     "@types/sinon": "^4.3.1",
-    "@typescript-eslint/eslint-plugin": "^2.4.0",
-    "@typescript-eslint/parser": "^2.4.0",
+    "@typescript-eslint/eslint-plugin": "~4.22.0",
+    "@typescript-eslint/parser": "~4.22.0",
     "babel-polyfill": "^6.13.0",
     "chai": "^4.1.2",
     "conventional-changelog-cli": "1.3.3",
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -66,16 +66,16 @@
...
     "redux": ">=4 <5",
-    "rxjs": ">=6.0.0-beta.0 <7",
-    "tslib": "^1.9.0"
+    "rxjs": "^7.0.0",
+    "tslib": "~2.1.0"
   },
   "devDependencies": {
     "@types/chai": "^3.5.2",
     "@types/mocha": "^2.2.48",
     "@types/node": "^12.7.12",
```
casoetan, line 75: The minimum node version is now 14.14.3
```
Symbol.observable: rxjs@7 is only compatible with @types/node@14.14.3 or higher and symbol-observable@3.0.0 and heigher. Older versions of @types/node incorrectly defined Symbol.observable and will be in conflict with rxjs and symbol-observable@3.0.0.
```

package.json
```diff
@@ -93,11 +93,11 @@
...
     "rimraf": "^2.5.4",
-    "rxjs": "^6.0.0",
+    "rxjs": "^7.0.0",
     "sinon": "^4.5.0",
     "ts-loader": "^6.2.0",
-    "tslib": "^1.10.0",
-    "typescript": "^3.6.4",
+    "tslib": "~2.1.0",
+    "typescript": "~4.2.2",
     "webpack": "^4.29.3",
```
casoetan, line 101: Should we look at updating webpack to 5?

src/createEpicMiddleware.ts
```diff
@@ -30,7 +31,7 @@ export function createEpicMiddleware<
   // other RxJS code outside of redux-observable internals.
   const QueueScheduler: any = queueScheduler.constructor;
   const uniqueQueueScheduler: typeof queueScheduler = new QueueScheduler(
-    (queueScheduler as any).SchedulerAction
+    (queueScheduler as any).SchedulerAction || (queueScheduler as any).schedulerActionCtor
```
casoetan, line 34: Should this just be `(queueScheduler as any).schedulerActionCtor`? I don't think this release would be backward compatible, so there may be no need to keep `ScheduleAction`

casoetan (commented):
Just some minor comments.

casoetan:
Thanks, @eltiodelsofa for taking this on. Looking to update as soon as possible 😄

package.json
evertbouw: Perhaps, but out of scope for this PR

src/createEpicMiddleware.ts
evertbouw: Since we now require Rxjs 7, it's fine to break compatibility with 6.

evertbouw (changes requested):
if you bump @types/node and remove the old `SchedulerAction` I think this is ready

.eslintrc
```diff
@@ -257,6 +257,11 @@
     "@typescript-eslint/unbound-method": "off",
     "@typescript-eslint/ban-ts-ignore": "off",
     "@typescript-eslint/no-empty-function": "off",
+    "@typescript-eslint/no-unsafe-assignment": "off",
+    "@typescript-eslint/no-unsafe-call": "off",
+    "@typescript-eslint/no-unsafe-member-access": "off",
+    "@typescript-eslint/no-unsafe-return": "off",
+    "@typescript-eslint/restrict-plus-operands": "off",
```
MaximeBernard, line 264: Do we really need to disable these rules?

package.json
eltiodelsofa: thanks for realizing this

casoetan (approved):
LGTM

evertbouw (approved):

PR #748, opened 2021-05-10
fix(dependencies): rxjs and tslib are now listed as dependencies

evertbouw:
We directly import from these libraries, so they should be in dependencies. This should prevent issues with plug'n'play in Yarn2. Redux is only used for type imports so it can stay in peerDependencies.

In #740 I said we could disable tslib, I'm not sure how I reached that conclusion because it's being used for StateObservable.

closes https://github.com/redux-observable/redux-observable/issues/740

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,6 +7,10 @@
     "": {
       "version": "2.0.0-rc.1",
       "license": "MIT",
+      "dependencies": {
+        "rxjs": "^7.0.0",
+        "tslib": "~2.1.0"
+      },
       "devDependencies": {
         "@types/chai": "^3.5.2",
         "@types/mocha": "^2.2.48",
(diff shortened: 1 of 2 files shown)


evertbouw:
if I build the project right now it produces this file

StateObservable.js
```js
import { __extends } from "tslib";
import { Observable, Subject } from 'rxjs';
var StateObservable = /** @class */ (function (_super) {
    __extends(StateObservable, _super);
    function StateObservable(input$, initialState) {
        var _this = _super.call(this, function (subscriber) {
            var subscription = _this.__notifier.subscribe(subscriber);
            if (subscription && !subscription.closed) {
                subscriber.next(_this.value);
            }
            return subscription;
        }) || this;
        _this.__notifier = new Subject();
        _this.value = initialState;
        input$.subscribe(function (value) {
            // We only want to update state$ if it has actually changed since
            // redux requires reducers use immutability patterns.
            // This is basically what distinctUntilChanged() does but it's so simple
            // we don't need to pull that code in
            if (value !== _this.value) {
                _this.value = value;
                _this.__notifier.next(value);
            }
        });
        return _this;
    }
    return StateObservable;
}(Observable));
export { StateObservable };
```

evertbouw:
I'm not aware of any other issues, should be good to go @jayphelps

PR #785, opened 2023-11-20
DRAFT: Redux5

evertbouw:
Should we remove support for action.type.toString()?
Webpack probably needs an upgrade or is it time to replace it?


evertbouw:
feel free to ignore this pr if you'd want a different approach

jayphelps:
@evertbouw Thank you much for this!

> Should we remove support for action.type.toString()?

I think we'd want to keep support for older versions of Redux, where folks might still be having types that are not strings e.g. symbols. What do you think?

>  Should we mark ofType as deprecated and recommend filter(action.match)?

jayphelps:
@evertbouw Thank you much for this!

> Should we remove support for action.type.toString()?

I think we'd want to keep support for older versions of Redux, where folks might still be having types that are not strings e.g. symbols. What do you think?

>  Should we mark ofType as deprecated and recommend filter(action.match)?

Similar to the previous argument that older versions of Redux or those not using the Redux Toolkit, having `ofType` is helpful for them. That said, I do like the idea of recommending `filter(action.match)` as the preferred approach moving forward. That would likely mean updating the usage docs, and probably adding a comment to the `ofType` docs mentioning it's there for legacy use. What do you think, though?

evertbouw:
Action types must be strings according to redux 5 core https://deploy-preview-3089--redux-starter-kit-docs.netlify.app/migrations/migrating-1.x-to-2.x#action-types-must-be-strings. I don't think we need the upcoming release to support both redux 4.x and 5.x.

Recommending `filter(action.match)` would be mostly in the docs and/or with deprecated hints yes.

PR #795, opened 2024-01-11
Port tests to vitest

evertbouw:
~~Only issue right now is that there are 2 tests for dealing with `uncaughtException` that I can't get vite to ignore~~
fixed by using `config.onUnhandledError`


jayphelps:
For unrelated reasons could you email me? Email in profile.

MaximeBernard (approved):
LGTM and thank you so much

